PR #11, opened 2024-01-31
Fix maxSupply and totalSupply issues

befitsandpiper:
- Flipping maxSupply and maxCapacity chain logic
- totalSupply subtract burns from total supply for error call(this probably won't change anything because mints doesn't seem to have access to these new values, but it can't hurt)
- tests continue to pass as expected


herr-bubu:
To fix this `oclif.manifest` needs to be changed to show a proper version(`1.0.8`)

PR #12, opened 2024-02-01
 TokenBalance supports TokenHold locks for fungible tokens

sentientforest:
TokenBalance supports locks on fungible tokens and accounts for locked fungible quantity when calculating spendable balance. 

LockToken / UnlockToken methods support new locking api for fungible tokens on TokenBalance.

TokenBalance can possibly support multiple locks of varying quantities over the total overall fungible balance, and will unlock quantity using the holds that are set to expire soonest.

### DIFF
--- a/chain-api/src/types/TokenBalance.spec.ts
+++ b/chain-api/src/types/TokenBalance.spec.ts
@@ -29,16 +29,11 @@ function emptyBalance() {
   });
 }
 
-function createHold(
-  instance: BigNumber,
-  expires: number,
-  quantity?: BigNumber | undefined,
-  name?: string | undefined
-) {
+function createHold(instance: BigNumber, expires: number, quantity?: BigNumber, name?: string) {
   return new TokenHold({
     createdBy: "user1",
     instanceId: instance,
-    quantity: quantity ?? new BigNumber("1"),
+    quantity: quantity ?? new BigNumber(1),
     created: 1,
     expires: expires,
     name: name
@@ -129,13 +124,13 @@ describe("fungible", () => {
   it("should fail to subtract quantity if quantity is locked by TokenHold", () => {
     // Given
     const balance = emptyBalance();
-    const hold = createHold(TokenInstance.FUNGIBLE_TOKEN_INSTANCE, 0, new BigNumber("10"));
+    const hold = createHold(TokenInstance.FUNGIBLE_TOKEN_INSTANCE, 0, new BigNumber(10));
 
-    balance.ensureCanAddQuantity(new BigNumber("10")).add();
+    balance.ensureCanAddQuantity(new BigNumber(10)).add();
     balance.ensureCanLockQuantity(hold).lock();
 
     // When
-    const error = () => balance.ensureCanSubtractQuantity(new BigNumber("1"), Date.now());
+    const error = () => balance.ensureCanSubtractQuantity(new BigNumber(1), Date.now());
 
     // Then
     expect(error).toThrow("Insufficient balance");
@@ -143,9 +138,9 @@ describe("fungible", () => {
 
   it("should successfully subtract quantity if the total is only partially locked", () => {
     // Given
-    const initialTotal = new BigNumber("10");
-    const lockedTotal = new BigNumber("5");
-    const subtractTotal = new BigNumber("1");
+    const initialTotal = new BigNumber(10);
+    const lockedTotal = new BigNumber(5);
+    const subtractTotal = new BigNumber(1);
 
     const balance = emptyBalance();
     const hold = createHold(TokenInstance.FUNGIBLE_TOKEN_INSTANCE, 0, lockedTotal);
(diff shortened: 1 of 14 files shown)


chain-api/src/types/TokenBalance.ts
```diff
@@ -403,18 +430,107 @@ export class TokenBalance extends ChainObject {
...
+      if (b.expires === 0 && a.expires === 0) {
+        return 0;
+      } else if (b.expires === 0) {
+        return -1;
+      } else if (a.expires === 0 || a.expires > b.expires) {
+        return 1;
+      } else {
+        return -1;
+      }
+    });
```
dzikowski, line 487: I wonder if we can make a separate method for that like `getUnexpiredLockedHoldsWithExpirationAscending`. I know the name is ugly, but it puts the sorting logic outside of this method and makes it easier to understand the rest

chain-api/src/types/TokenBalance.ts
```diff
@@ -403,18 +430,107 @@ export class TokenBalance extends ChainObject {
...
+
+  private getLockedQuantity(lockedHolds: TokenHold[]) {
+    let lockedQuantity: BigNumber = new BigNumber("0");
+
+    for (const hold of lockedHolds) {
+      lockedQuantity = lockedQuantity.plus(hold.quantity);
+    }
+
+    return lockedQuantity;
+  }
```
dzikowski, line 532: what about more functional style?
```suggestion
  private getCurrentLockedQuantity(currentTime: number): BigNumber {
    return this.getUnexpiredLockedHolds(currentTime)
      .reduce((h, sum) => sum.plus(h.quantity), new BigNumber(0));
  }
```

chaincode/src/locks/lockTokens.ts
```diff
@@ -94,6 +88,13 @@ export async function lockToken(
     const msg = `LockToken executed on behalf of another user (fromPerson: ${owner}, callingUser: ${callingOnBehalf})`;
     ctx.logger.info(msg);
 
+    // for initial support of fungible tokens, require owner to be the caller.
+    if (tokenInstanceKey.isFungible()) {
+      throw new NotImplementedError("LockToken is not supported for fungible tokens", {
+        instanceKey: tokenInstanceKey.toStringKey()
+      });
+    }
```
dzikowski, line 96:
```suggestion
    const msg = `LockToken executed on behalf of another user (fromPerson: ${owner}, callingUser: ${callingOnBehalf})`;
    ctx.logger.info(msg);

    // for initial support of fungible tokens, require owner to be the caller.
    if (tokenInstanceKey.isFungible()) {
      throw new NotImplementedError(`${msg} is not supported for fungible tokens`, {
        instanceKey: tokenInstanceKey.toStringKey()
      });
    }
```

chaincode/src/locks/unlockToken.ts
```diff
@@ -24,16 +25,15 @@ import { UnlockForbiddenUserError } from "./LockError";
 export interface UnlockTokenParams {
   tokenInstanceKey: TokenInstanceKey;
   name: string | undefined;
+  quantity?: BigNumber | undefined;
```
dzikowski, line 28:
```suggestion
  quantity: BigNumber | undefined;
```

chain-api/src/types/TokenBalance.spec.ts
```diff
@@ -112,11 +120,145 @@ describe("fungible", () => {
...
+    // When
+    balance.ensureCanUnlockQuantity(quantityLockedPerHold, Date.now()).unlock();
+    const unexpiredLockedHolds = balance.getUnexpiredLockedHolds(Date.now());
+
+    // Then
+    expect(balance.getLockedQuantityTotal(Date.now())).toEqual(lockedTotal.minus(quantityLockedPerHold));
+    expect(unexpiredLockedHolds.length).toBe(2);
+    expect(unexpiredLockedHolds[0].expires).toBe(expiresIn30DaysTimestamp);
+    expect(unexpiredLockedHolds[1].expires).toBe(noExpirationTimestamp);
+  });
```
dzikowski, line 261: great tests! I think we can also add a test to verify partial unlock passes, like add two locks with quantity = 5, and unlock 7

dzikowski (approved):

chain-api/src/types/TokenBalance.spec.ts
sentientforest: I'm glad you recommended this test case, because while implementing I realized I'd failed to handle that scenario yesterday. Fixed in latest commit.

chain-api/src/types/TokenBalance.ts
sentientforest: This is a good call, I ended up moving the sort logic itself to a static method on the `TokenHold` class (it may be useful down the line for other methods / places in the code), and added `getUnexpiredLockedHoldsSortedByAscendingExpiration` as a method to get the sorted values.

Jehosephat (approved):

PR #25, opened 2024-02-08
flip supply and capacity logic

befitsandpiper:
Flipping maxSupply and maxCapacity chain logic


chaincode/src/allowances/grantAllowance.ts
```diff
@@ -234,7 +234,7 @@ async function grantAllowanceByPartialKey(
 export function ensureQuantityCanBeMinted(
   tokenClass: TokenClass,
   quantity: BigNumber,
-  totalKnownMintsCount?: BigNumber,
+  totalKnownMintAllowanceCount?: BigNumber,
   totalKnownBurnsCount?: BigNumber
```
Jehosephat, line 238: Should this method be taking into account both knownMints and knownAllowances against the maxCapacity?

befitsandpiper: I don't think so:
- for maxSupply we already account for burns
- for maxCapacity we do not account for burns bc we only care about total minted ever, but there shouldn't be the ability to grant mint allowances greater than the capacity I would think...

also, regardless, this is really just a renaming to account for the reality, which is is that this value is always coming in as the mint allowance count except in the specific case where the mint allowance has already been granted and you are simply using that allowance.

The renaming here doesn't change the underlying functionality, but is just for clarity.

Jehosephat: Ok, I think that's ok since it keeps the status quo, but I am concerned that this would not prevent granting allowances beyond the maxCapacity that then could not be minted.

befitsandpiper: That is true... it will not allow granting allowances past capacity if they are already granted. It's something we can revisit when we are removing mintToken and updating batchMint

Jehosephat: I should be more specific about the scenario. If I have been granted an allowance for tokens, and have not minted it and in the meantime enough tokens are minted to hit maxCapacity then my allowance is now invalid, effectively, right? If my allowance is not being accounted for when figuring out if tokens can be minted under maxCapacity, then this scenario seems possible.

befitsandpiper: The allowance is accounted for... because we're using total allowances for the check instead of mints. The scenario you're talking about would happen if we specifically checked against mints and not allowances.

Jehosephat (approved):

PR #41, opened 2024-02-18
Fix #38 : Redundant $ in error messages

### LINKED ISSUE #38: Redundant $ in error messages
There are some parts in the code where we have redundant `$` in error message, for instance:

```
super(`BurnTokens failed: ${message}$`, payload);
```

We should scan the code and remove the redundant character.

MohitKambli:
Hello,
Hope you're doing well.
According to problem statement of issue #38, as a part of fix, I have removed the redundant $ sign from the following three files where the error message is being displayed:

1. chaincode/src/burns/BurnError.ts
2. chaincode/src/locks/LockError.ts
3. chaincode/src/transfer/TransferError.ts

The code change looks something like this:
`From ${message}$ to ${message}`

I tried testing the code changes, but for some reason I wasn't able to successfully execute some commands as mentioned in the [Getting Started](https://github.com/GalaChain/sdk/blob/main/docs/getting-started.md#getting-started) guide.
The error looks something like this:

image

If at all there are any more changes to be made or anything else is required from my end, then do let me know.
Sincere apologies if you find any mistakes implemented from my end.

Thanks and Regards,
Mohit Kambli


dzikowski:
Thank's again for the contribution. The change is fairly simple, so our CI tests should be enough. However, if you want to run tests locally, using `network:start` script, and you want the tests to use the updated version of SDK that has been updated locally, the procesure is a bit complex.

It seems you try to execute `network:start` script from the root directory of the SDK. However, the script is included in `package.json` in `chain-cli/chaincode-template`, and you should call it from this directory. Also, if you want to use locally changed version of the SDK, you should call `../../npm-pack-and-replace.sh` script to make the template use your local changes (don't commit `package.json` and `package-lock.json` after that change).

So:
1. navigate in terminal to `chain-cli/chaincode-template`
2. execute `../../npm-pack-and-replace.sh`
3. execute `npm run network:start`

dzikowski (approved):

MohitKambli:
Hello @dzikowski,
Hope you're doing well.
As you had mentioned in the previous comment, I implemented the 3 steps.
In the second step, there were couple of commands getting executed, wherein I was able to see some error messages which looked something like this:

image

Furthermore, for the third step, here is what I am able to see:

image

If at all I am doing anything wrong or missing any steps, then do let me know.
If there are any other steps that I need to execute, then kindly let me know.
Sincere apologies for the trouble being caused.

Thanks and Regards,
Mohit Kambli

Fuerback:
hey @MohitKambli,

I noticed you are trying to execute the commands using a Windows terminal, but some of them don't work for Windows.
I'd suggest running it using WSL, you can follow our documentation [here](https://galahackathon.com/v1.0.14/getting-started/)

Issue #45, opened 2024-02-20
Misuse of getOrCreateFabloRoot


dzikowski:
`getOrCreateFabloRoot` is used in `network-prune/index.ts` in CLI project, but it may lead to creation of unwanted directories. If the directory is missing, instead of creating the directory, CLI should emit a warning and exit 0 (we want the user to see there is no such directory, but we also want to use it together with `up`, like `prune && up`, and we don't want the whole command chain fail because of missing directory)

PR #51, opened 2024-02-21
Verify if directory exists and log it

Fuerback:
This PR verifies if fablo root directory exists and log it instead of creating one.

Closes https://github.com/GalaChain/sdk/issues/45


chain-cli/src/commands/network-prune/index.ts
```diff
@@ -56,8 +60,11 @@ function downBrowserApi(fabloRoot: string): void {
...
-  const fabloRoot = path.resolve(fabloDir);
-  execSync(`mkdir -p "${fabloRoot}"`);
-  return fabloRoot;
+function getFabloRoot(fabloDir: string): string {
+  if (fs.existsSync(fabloDir)) {
+    return path.resolve(fabloDir);
+  } else {
+    console.log(`Directory '${fabloDir}' does not exist.`);
+    return "";
+  }
```
dzikowski, line 69: This way maybe? To have it more explicit?

```suggestion
function getFabloRoot(fabloDir: string): string | undefined {
  if (fs.existsSync(fabloDir)) {
    return path.resolve(fabloDir);
  } else {
    console.warn(`Directory '${fabloDir}' does not exist.`);
    return undefined;
  }
```

herr-bubu (approved):

PR #64, opened 2024-02-26
Fix ci jobs when unify version

Fuerback:
This PR fixes the `Chaincode Template Lint` job when `unifyVersions.js` is executed.

It also updates the `unifyVersions.js` to update a few files (e.g. licenses, lock, oclif manifest) after update the `package.json` files.


.github/workflows/ci.yml
```diff
@@ -1,7 +1,6 @@
 name: CI
 on:
   push:
-    branches: ['**']
```
dzikowski, line 4: is it safe to remove it?

.github/workflows/ci.yml
```diff
@@ -122,65 +121,63 @@ jobs:
...
+        run: |
+          $package = Get-Content -Path test-project/package.json -Raw
+          if ($package -match '@gala-chain/test-project') {
+            Write-Host "package.json contains '@gala-chain/test-project'"
+          } else {
+            Write-Host "package.json does not contain '@gala-chain/test-project'"
+            exit 1
+          }
 
-  # chaincode-template-lint:
```
dzikowski, line 156: comment to be removed

dzikowski (approved):

.github/workflows/ci.yml
Fuerback: It's duplicating the job executions after we added the `pull_request`.
My understanding is that we should choose which one we want to keep, for every branch (and it includes pull requests) or only when a pull request is created.

.github/workflows/ci.yml
Fuerback: There is no more comment line in ci.yml.

herr-bubu (approved):

PR #68, opened 2024-02-28
Generate access keys during init command

### LINKED ISSUE #21: Proposal: Generate key pair along with initialization from template
You can create a chaincode from template using:

```
galachain init my-gc-chaincode
```

However, in order to connect and deploy to testnet, you also need to have a secp key pair, which can be generated with the command:

```
mkdir my-gc-chaincode/keys
galachain keygen my-gc-chaincode/keys/gc-admin-key
galachain keygen my-gc-chaincode/keys/gc-dev-key
```

which creates a secp key pair (in this case `gc-key` and `gc-key.pub`).

I think we can extend our `galachain init` command to:
- [ ] generate key pairs for `admin` and `dev1` in `my-gc-chaincode/keys` (default names, no parametrization)
- [ ] add private keys to `.gitignore`
- [ ] add `gc-dev-key` to `.gitignore` - since it is related with local env

And additionally:
- [ ] make the default path for `gc-dev-key` a default key for signing in all commands

This way we achieve consistent way of managing the keys.

herr-bubu:
Closes #21


chain-cli/src/keys.ts
```diff
@@ -30,11 +30,7 @@ export async function generateKeys(keysPath: string): Promise<void> {
   const devPrivateKey = secp.utils.bytesToHex(secp.utils.randomPrivateKey());
   const devPublicKey = secp.utils.bytesToHex(secp.getPublicKey(adminPrivateKey));
 
-  if (process.platform === "win32") {
-    execSync(`mkdir ${keysPath}`);
-  } else {
-    execSync(`mkdir -p ${keysPath}`);
-  }
+  execSync(`mkdir -p ${keysPath}`);
```
herr-bubu, line 33: I deliberately removed Windows support as it won't be neccessary after this one is merged: https://github.com/GalaChain/sdk/pull/67

dzikowski (approved):

chain-cli/src/commands/connect/index.ts
```diff
@@ -45,7 +46,7 @@ export default class Connect extends BaseCommand<typeof Connect> {
   async run(): Promise<void> {
     const { args, flags } = await this.parse(Connect);
 
-    const developerPrivateKey = args.developerPrivateKey ?? process.env.DEV_PRIVATE_KEY;
+    const developerPrivateKey = args.developerPrivateKey ?? (await getPrivateKey());
```
dzikowski, line 49:
```suggestion
    const developerPrivateKey = await getPrivateKey(args.developerPrivateKey);
```

```
getPrivateKey(keyArg: string | undefined)
```

chain-cli/src/commands/info/index.spec.ts
```diff
@@ -69,6 +71,8 @@ describe("ChainInfo Command", () => {
 
     process.env = { ...process.env, DEV_PRIVATE_KEY: undefined };
 
+    jest.spyOn(fs, "readFileSync").mockReturnValue("");
```
dzikowski, line 74: to be verified if missing file throws an exception or returns empty string

chain-cli/src/galachain-utils.ts
```diff
@@ -146,9 +139,35 @@ export async function deployChaincode(params: {
...
 }
 
+export async function generateKeys(keysPath: string): Promise<void> {
+  const adminPrivateKey = secp.utils.bytesToHex(secp.utils.randomPrivateKey());
+  const adminPublicKey = secp.utils.bytesToHex(secp.getPublicKey(adminPrivateKey));
+
+  const devPrivateKey = secp.utils.bytesToHex(secp.utils.randomPrivateKey());
+  const devPublicKey = secp.utils.bytesToHex(secp.getPublicKey(adminPrivateKey));
+
+  execSync(`mkdir -p ${keysPath}`);
```
dzikowski, line 149: use `fs.mkdir()`

dzikowski (approved):

chain-cli/src/commands/connect/index.ts
```diff
@@ -45,7 +46,7 @@ export default class Connect extends BaseCommand<typeof Connect> {
   async run(): Promise<void> {
     const { args, flags } = await this.parse(Connect);
 
-    const developerPrivateKey = args.developerPrivateKey ?? process.env.DEV_PRIVATE_KEY;
+    const developerPrivateKey = args.developerPrivateKey ?? (await getPrivateKey(args.developerPrivateKey));
```
dzikowski, line 49:
```suggestion
    const developerPrivateKey = await getPrivateKey(args.developerPrivateKey);
```

dzikowski (approved):

PR #77, opened 2024-03-05
Automated Version bump

herr-bubu:


.github/workflows/ci.yml
```diff
@@ -148,11 +148,70 @@ jobs:
...
+          branch: version-bump-"${{ env.NEW_SDK_VERSION }}"
+          delete-branch: true
+          body: "Version bump to ${{ env.NEW_SDK_VERSION }}"
+          labels: version-bump
+          token: ${{ secrets.GITHUB_TOKEN }}
+  auto-merge:
+    needs: [ bump-version ]
+    name: Auto merge
+    runs-on: ubuntu-22.04
+    if: github.event.pull_request.user.login == 'gala-user[bot]'
```
herr-bubu, line 204: maybe we don't need this check here. We could assume that if there's a new tag, everything is fine so the bump PR can be merged automatically without even checking who did that

dzikowski (approved):

Issue #100, opened 2024-03-15
Replace 'mainnet' for 'sandbox' in CLI


Fuerback:
Replace 'mainnet' for 'sandbox' in CLI

PR #101, opened 2024-03-15
Replace 'mainnet' for 'sandbox'

Fuerback:
This PR replaces the word 'mainnet' for 'sandbox' in CLI.

Closes https://github.com/GalaChain/sdk/issues/100


chain-cli/src/commands/deploy/index.ts
```diff
@@ -52,7 +52,7 @@ export default class Deploy extends BaseCommand<typeof Deploy> {
   async run(): Promise<void> {
     const { args, flags } = await this.parse(Deploy);
 
-    const environment = flags.testnet ? "testnet" : "mainnet";
+    const environment = flags.testnet ? "testnet" : "sandbox";
```
dzikowski, line 55: double check if it does not break anything

dzikowski (approved):

chain-cli/src/commands/deploy/index.ts
Fuerback: all good, this name is not used in the code, it's just logged to show the user which environment he wants to deploy
the deploy environment is chosen by the boolean `flags.testnet`

PR #107, opened 2024-03-18
Update getting-started.md

IndiaJonathan:


Fuerback (approved):

docs/getting-started.md
```diff
@@ -39,6 +39,11 @@ Install all dependencies:
 npm i
 ```
 
+Build the project:
+
+```
+npm run build
+```
```
dzikowski, line 46: Actually this step is redundant, because `npm run network:start` also runs build in watch mode. But this is actually a good catch, perhaps we should be more explicit on this in the next section, like:

> The network will build the chaincode and start in hot-reload/watch mode, so leave the prompt with logs running and execute the following commands in a new prompt.

PR #109, opened 2024-03-19
Adding rest confs by default

jburk-gala:


chain-cli/network/fablo-config-dev-mode.json
```diff
@@ -21,7 +21,8 @@
           "type": "solo",
           "instances": 1
         }
-      ]
+      ],
+      "tools": { "fabloRest": true }
```
dzikowski, line 25: yeah, I've tried that before. The case is in dev/watch mode Fabric discovery service does not see deployed chaincodes. And Fablo REST relies on service discovery for connecting to Fabric network. So Fablo REST won't work in dev mode - it won't be able to call chaincodes. If we want to make it work, we need to change Fablo REST source code.

chain-cli/network/fablo-config-default.json
```diff
@@ -20,7 +20,8 @@
           "type": "raft",
           "instances": 1
         }
-      ]
+      ],
+      "tools": { "fabloRest": true }
```
dzikowski, line 24: This is not required for Orderer organization, since technically you are not supposed to call chaincodes from orderer organization

dzikowski:
closing, because this is not feasible with our local setup

PR #114, opened 2024-03-21
Set up docker buildx

Fuerback:
This PR sets up the Docker Buildx to build and publish the docker images for all platforms.


herr-bubu (approved):

.github/workflows/publish-on-tag.yml
```diff
@@ -131,8 +131,10 @@ jobs:
       contents: read
       packages: write
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
```
dzikowski, line 135: will it work without checkout?

Fuerback: I've read somewhere the checkout step wasn't necessary because `docker/setup-buildx-action@v3` does that, but I'm not finding this information anymore and the example there uses checkout.
Going to restore it.

Issue #180, opened 2024-04-15
Verify docker image architecture in CLI


dzikowski:
CLI calls docker image to get contract list. At the same time it should verify if the architecture of docker image is valid (it should be linux/amd64, arm is not supported)

PR #185, opened 2024-04-16
Check Docker image architecture before run it

Fuerback:
This PR check the Docker image architecture before run it.

Closes https://github.com/GalaChain/sdk/issues/180


chain-cli/src/galachain-utils.ts
```diff
@@ -99,6 +99,15 @@ export async function getDeploymentResponse(params: { privateKey: string; isTest
 }
 
 function getContractNames(imageTag: string): { contractName: string }[] {
+  const imageArchitecture = execSync(
+    "docker inspect --format=json ${imageTag} | jq -r '.[0].Os + \"/\" + .[0].Architecture'"
+  )
+    .toString()
+    .trim();
```
dzikowski, line 106: if it responds with JSON, then I think the better way is to get the command result, parse JSON in TS, and take proper value. This way we:
1. minimize the number of commands we call
2. minimize the usage of jq (which is still non-standard library; and we can imagine someone wants to deploy without running locally, where jq is required)

dzikowski (approved):

PR #211, opened 2024-04-30
Support MetaMask signatures

befitsandpiper:
Add prefix field to dtos to allow compatibility with ethereum personal sign and eventually ethereum signing v4.
Also includes unit test with signature and publicKey recovery from metamask signing test case.


chain-api/src/utils/signatures.spec.ts
```diff
@@ -315,4 +315,20 @@ describe("signatures", () => {
...
+    
+    // When
+    const metamaskPubKey = signatures.recoverPublicKey(metamask, metamaskPayload, "\u0019Ethereum Signed Message:\n" + JSON.stringify(metamaskPayload).length);
+    const galachainPubKey = signatures.recoverPublicKey(galachain, galachainPayload);
+  
+    // Then
+    console.log('metamaskPubKey', metamaskPubKey, "galachainPubKey", galachainPubKey)
+    expect(metamaskPubKey).toEqual(galachainPubKey)
+  });
 });
```
dzikowski, line 334: I suggest to rework a little this test, to make it even more clear:

1. Missing `// Given` comment on the top
2. Add `const metamaskPrefix = "\u0019Ethereum Signed Message:\n" + signatures.getPayloadToSign(metamaskPayload).length`. (Note we don't use `JSON.stringify` here, because for some payloads it may produce different input than our way of serializing data)
3. `signature` from `galaChainPayload` may be removed, it's ignored.
4. It would be good to add a comment pointing that `uniqueKey` is different for each payload (this is the only difference, which is important, but also easy to miss)
5. Redundant `console.log`
6. Also please remove redundant quotes from payload field names (is linter not complaining?)

chain-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gala-chain/api",
-  "version": "1.1.16",
+  "version": "1.1.17",
```
dzikowski, line 3: you don't need to update versions anymore in code

chain-api/src/types/dtos.ts
```diff
@@ -133,6 +133,15 @@ export class ChainCallDTO {
   @IsNotEmpty()
   public signature?: string;
 
+  @JSONSchema({
+    description:
+      "Prefix for metamask transaction signatures. " +
```
dzikowski, line 138:
```suggestion
      "Prefix for MetaMask transaction signatures. " +
```

dzikowski (approved):

Issue #213, opened 2024-05-01
BurnTokens Does Not Use Up Burn Allowances


Fuerback:
When burning tokens, the usesSpent and quantitySpent on the burn allowances do not seem to be updating.

Please see attached before, burn and after (staging) requests and responses.

PR #214, opened 2024-05-01
Filter allowances when using allowances

Fuerback:
This PR adds a filter on `useAllowances` function to make sure the AllowanceType used is the required one.

Closes https://github.com/GalaChain/sdk/issues/213


chaincode/src/burns/burnTokens.ts
Jehosephat, line 1: I think it may still make sense to send in the AllowanceType.Burn here and have the applicableAllowanceResponse only consist of burn allowances. The change to useAllowances to take a type is a good general failsafe.

befitsandpiper (approved):
looks good to me, good job

Jehosephat (approved):

PR #226, opened 2024-05-14
Update Fabric libs to 2.5.4 and resolve conflicting grpc-js dependency

dzikowski:


dzikowski:
It turns out `fabric-shim` (lib for chaincodes) used `@grpc/grpc-js` in version `1.8.15` which needs to be enforced for chaincodes. We also use `fabric-network` (chaincode client library), which, depending on the version uses `@grpc/grpc-js` in `~1.9.x` or `~1.7.x`, so there is no overlap. Thus, I had to enforce `@grpc/grpc-js` for `1.8.15` in `@gala-chain/chaincode` and for chaincode template in `@gala-chain/cli`.

herr-bubu (approved):
🎉

PR #230, opened 2024-05-15
Allow compact public key for admin env

dzikowski:


chaincode/src/services/PublicKeyService.ts
```diff
@@ -79,7 +79,8 @@ export class PublicKeyService {
 
     // check if we want the profile of the admin
     if (process.env.DEV_ADMIN_PUBLIC_KEY !== undefined && process.env.DEV_ADMIN_USER_ID !== undefined) {
-      const adminEthAddress = signatures.getEthAddress(process.env.DEV_ADMIN_PUBLIC_KEY);
+      const nonCompactPK = signatures.getNonCompactHexPublicKey(process.env.DEV_ADMIN_PUBLIC_KEY);
+      const adminEthAddress = signatures.getEthAddress(nonCompactPK);
```
Rasbandit, line 83: Would it make more sense here to try and cover both cases. Like having logic around seeing if it needs to be decompressed.

dzikowski: Actually everything happens inside `getNonCompactHexPublicKey`. When it gets non-compact key, it returns the same value. Does it answer your question? Because I'm not sure if I fully understood

Rasbandit: Yup, thats exactly it. Thanks

chaincode/src/services/PublicKeyService.ts
```diff
@@ -79,7 +79,8 @@ export class PublicKeyService {
 
     // check if we want the profile of the admin
     if (process.env.DEV_ADMIN_PUBLIC_KEY !== undefined && process.env.DEV_ADMIN_USER_ID !== undefined) {
```
Rasbandit, line 81: Could we have to check for falsy values instead of undefined? In our deploy pipeline we have to include the DEV_ADMIN_PUBLIC_KEY and DEV_ADMIN_USER_ID values, but if it is not set then it makes it an empty string. Internally we ran into an issue where because the values were `""` and not undefined it triggered this block and blocked users from being registered with basic-asset.

Rasbandit, line 81:
```suggestion
    if (process.env.DEV_ADMIN_PUBLIC_KEY && process.env.DEV_ADMIN_USER_ID) {
```

Issue #240, opened 2024-05-30
checkAllowances() erroneously filters out allowances that have not expired


sentientforest:
The `checkAllowances()` function defined in `chaincode/src/allowances/checkAllowances/ts` contains a bug that filters out allowances that have not expired, instead of allowances that have expired. 

https://github.com/GalaChain/sdk/blob/main/chaincode/src/allowances/checkAllowances.ts#L114

```
      (allowance.expires === 0 || isAllowanceExpired(ctx, allowance))
```

should be

```
      (allowance.expires === 0 || !isAllowanceExpired(ctx, allowance))
```

We'll want some unit tests to replicate the issue and a patch to fix it.

PR #241, opened 2024-05-30
fix: checkAllowances includes allowances that are not yet expired

sentientforest:
Fix issue 
#240

### DIFF
--- a/chaincode/src/allowances/checkAllowances.ts
+++ b/chaincode/src/allowances/checkAllowances.ts
@@ -111,7 +111,7 @@ export async function checkAllowances(
       allowance.additionalKey === tokenInstanceKey.additionalKey &&
       allowance.instance.isEqualTo(tokenInstanceKey.instance) &&
       allowance.allowanceType === action &&
-      (allowance.expires === 0 || isAllowanceExpired(ctx, allowance))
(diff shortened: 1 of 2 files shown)


Fuerback (approved):

chaincode/src/allowances/checkAllowances.ts
```diff
@@ -111,7 +111,7 @@ export async function checkAllowances(
       allowance.additionalKey === tokenInstanceKey.additionalKey &&
       allowance.instance.isEqualTo(tokenInstanceKey.instance) &&
       allowance.allowanceType === action &&
-      (allowance.expires === 0 || isAllowanceExpired(ctx, allowance))
+      (allowance.expires === 0 || !isAllowanceExpired(ctx, allowance))
```
mistval, line 114: I notice the `isAllowanceExpired()` function returns false if expires is zero, so I suppose this line could just be `!isAllowanceExpired(ctx, allowance)` without checking the zero case here.

mistval (approved):

chaincode/src/allowances/checkAllowances.ts
sentientforest: Good call, simpler is better. I pushed a small update

PR #248, opened 2024-06-05
Add type safe clients for PublicKeyContract and TokenContract in connect library

befitsandpiper:
add endpoints to public key and token client objects, add setters for url, make submit function public so clients can use custom signing


chain-connect/src/GalachainConnectClient.ts
```diff
@@ -103,3 +148,145 @@ export class GalachainConnectClient {
     return this.calculatePersonalSignPrefix(newPayload);
   }
 }
+
+export class PublicKeyClient {
```
dzikowski, line 152: I think both `PublicKeyClient` and `TokenClient` should be in separate files

chain-connect/src/GalachainConnectClient.ts
```diff
@@ -103,3 +148,145 @@ export class GalachainConnectClient {
...
+  constructor(
+    private client: GalachainConnectClient,
+    private url: string
+  ) {}
+
+  public updateUrl(url: string) {
+    this.url = url;
+  }
```
dzikowski, line 160:
```suggestion
  constructor(
    private readonly client: GalachainConnectClient,
    private readonly url: string
  ) {}
```

I think it is better to keep it simple, with no mutations.

chain-connect/src/GalachainConnectClient.ts
```diff
@@ -103,3 +148,145 @@ export class GalachainConnectClient {
...
+
+export class TokenClient {
+  constructor(
+    private client: GalachainConnectClient,
+    private url: string
+  ) {}
+
+  public updateUrl(url: string) {
+    this.url = url;
+  }
```
dzikowski, line 187:
```suggestion
  constructor(
    private readonly client: GalachainConnectClient,
    private readonly url: string
  ) {}
```

dzikowski (approved):

PR #258, opened 2024-06-11
Fix potential ECDSA signature malleability - low S value

befitsandpiper:
add malleability check, flip signatures on signing in they don't meet the requirement, updates to tests


chain-api/src/utils/signatures.spec.ts
```diff
@@ -358,4 +378,29 @@ describe("signatures", () => {
...
+      flippedSignatureObj.r.toString("hex", 32) +
+      flippedSignatureObj.s.toString("hex", 32) +
+      new BN(flippedSignatureObj.recoveryParam === 1 ? 28 : 27).toString("hex", 1);
+
+    // When
+    const originalPubKey = recoverPublicKeyTestFunction(originalSig, payload);
+    const newPubKey = recoverPublicKeyTestFunction(flippedSig, payload);
+
+    // Then
+    expect(originalPubKey).toEqual(newPubKey);
```
dzikowski, line 404: a little suggestion on comments and the location of `originalPubKey` - the this we are testing here is flipping. It's a bit unreadable in the following diff, but the change is just to reorder some lines and comments.

```suggestion
    const originalPubKey = recoverPublicKeyTestFunction(originalSig, payload);

    // When
    const flippedSignatureObj = flipSignatureParity(normalizeSecp256k1Signature(originalSig));

    // Then
    const flippedSig =
      flippedSignatureObj.r.toString("hex", 32) +
      flippedSignatureObj.s.toString("hex", 32) +
      new BN(flippedSignatureObj.recoveryParam === 1 ? 28 : 27).toString("hex", 1);
    const newPubKey = recoverPublicKeyTestFunction(flippedSig, payload);

    expect(originalPubKey).toEqual(newPubKey);
```

chain-api/src/utils/signatures.ts
```diff
@@ -247,6 +276,13 @@ function signSecp256k1(dataHash: Buffer, privateKey: Buffer, useDer?: "DER"): st
 
   const signature = ecSecp256k1.sign(dataHash, privateKey);
 
+  // Low-S normalization
+  let newSig;
+  if (signature.s.cmp(ecSecp256k1.curve.n.shrn(1)) > 0) {
+    newSig = flipSignatureParity(signature);
+    newSig.toDER = signature.toDER;
+  }
+
```
dzikowski, line 285: Was the implementation complete here? It seems nothing happens with `newSig`

```suggestion
  let signature = ecSecp256k1.sign(dataHash, privateKey);

  if (signature.s.cmp(ecSecp256k1.curve.n.shrn(1)) > 0) {
    signature = flipSignatureParity(signature);
  }

```

befitsandpiper: The reason this is done this way is bc the newSig can be an Secp256k1Signature instead of EC.Signature, which doesn't have the toDER function on it. Maybe there's a better way of dealing with this type discrepancy

dzikowski (approved):

PR #264, opened 2024-06-12
Dry run flag simulates chaincode without writes

sentientforest:
New feature proposal - add a `dryRun: true` property to any contract method input that inherits from the `ChainCallDTO` class. 

Setting this property to `true` will execute the full chaincode lifecycle up until the `GalaChainStubCache` would normally `flushWrites()`. 

Instead of writing the results, return the writes, reads, and deletes as a new property on `GalaChainResponse`: `ReadWriteSet`.


Jehosephat (commented):
Given that one objective of fees is friction against griefing/spamming, the only concern that I have thought of with this feature is the possibility that the dryRun flag could allow a griefer to call a method in dryRun mode 'for free' and therefore have a vector for overwhelming the chain with traffic. But I suppose that's true of our non-fee-gated methods such as FetchBalances. So, I don't know that it changes anything, dryRun mode is basically an EVALUATE mode for a normally SUBMIT mode function.

dzikowski:
Proposed an alternative approach here: #281

sentientforest:
Nice, @dzikowski , looks to me like your alternative is simpler and should be easier to maintain going forward. 

Glad to see that we can use `evaluate` here in chaincode in lieu of modifying `submit` methods to forgo their writes when a flag is provided. 

Unless anyone else has comments, I think we should close out my pull request and go with your alternative.

dzikowski:
#281 merged, closing this one. Thanks @sentientforest!

PR #274, opened 2024-06-18
Increment patch verison for asset-chaincode using github actions

RhiannonNichols:
1. Github Action triggers Gitlab pipeline for asset-chaincode
2. Github Action Polls Gitlab for success, failure or cancelation of the job using the pipeline id.
3. If success, increment package.json with new patch version using semver.
4. If failed or canceled, we exit. Otherwise...
5. Pushes and tags

Please let me know your thoughts, fellas.


.github/workflows/trigger-gitlab-pipeline.yml
```diff
@@ -5,7 +5,40 @@ on:
...
+          response=$(curl --request POST \
+                           --form token=${{ secrets.GITLAB_TRIGGER_TOKEN }} \
+                           --form ref=main \
+                           "https://gitlab.com/api/v4/projects/35484791/trigger/pipeline")
+          echo "::set-output name=pipeline_id::$(echo $response | jq -r '.id')"
+      
+      - name: Poll Pipeline Status
+        run: |
+          status=""
+          while [[ "$status" != "success" && "$status" != "failed" ]]; do
```
Fuerback, line 26:
```suggestion
          while [[ "$status" != "success" ]]; do
```
we are checking if the status is failed inside the while, so I think it should work without this check here

.github/workflows/trigger-gitlab-pipeline.yml
```diff
@@ -5,7 +5,40 @@ on:
...
+            result=$(curl "https://gitlab.com/api/v4/projects/35484791/pipelines/${{ steps.trigger.outputs.pipeline_id }}")
+            status=$(echo $result | jq -r '.status')
+            echo "Pipeline status: $status"
+            if [[ "$status" == "failed" ]]; then
```
Fuerback, line 30: question: I don't know how many status this response might return, I guess is `success`, `pending` and `failed`, but I'd like to check if there is others that we might handle as an error as well, e.g. `cancelled`.

RhiannonNichols: Wonderful feedback!

Will validate states as I'm only aware of the three mentioned (success, failed, pending). It's very possible there could be additional, as you mentioned.

RhiannonNichols: Looks like these are the states a given pipeline in GitLab can reflect per https://docs.gitlab.com/ee/api/pipelines.html

created, waiting_for_resource, preparing, pending, running, success, failed, canceled, skipped, manual, scheduled

Fuerback: I thin should be enough to check `failed` and `canceled` here

RhiannonNichols: I went ahead and committed the changes we talked about. I think this covers it?

Fuerback (approved):

PR #281, opened 2024-06-24
Support DryRun method for dry run of any contract method

dzikowski:


chaincode/src/contracts/GalaContract.ts
```diff
@@ -92,42 +95,76 @@ export abstract class GalaContract extends Contract {
...
+      throw new NotFoundError(
+        `Method ${dto.method} not found. Available methods: ${availableMethods.join(", ")}`
+      );
+    }
+
+    // reset authorization so the method will be executed as the calling user from DTO
+    ctx.resetCallingUser();
+
+    // method needs to be executed first to populate reads, writes and deletes
+    const response = await this[methodName](ctx, dto.dto);
```
sentientforest, line 159: There's a lingering issue that I have been turning over still with this dry run feature. 

Both our approaches (this pull request and mine on [264](https://github.com/GalaChain/sdk/pull/264)) require a valid signature on the DTO from the end user, as I read it: I would expect from the above code that when the method is executed on line 159 the `GalaTransaction` decorator will validate the signature on the nested dto (`dto.dto`) and re-populate the `callingUser` based on that validation. 

However, I think it's a valid use case that an end user may not want to sign / authorize the transaction until after the results of the dry run are viewed and reviewed. 

In the above, I would expect that the signed dto stored in `dto.dto` could be used to submit the transaction. (If I'm reading that wrong, or missing something, do let me know).

So in order to execute a dry run, the end user is first providing a DTO that could be used to actually execute the method by some actor in the middle, or a relaying service of some kind. 

It might be a very unlikely scenario, but it could be possible that an end user signs a DTO expecting to review results of a dry run. Then, upon review, the end user decides not to go forward with the submission. But a malicious actor or bug in a relaying service submits the signed DTO anyway and the results get written to chain. It's an edge case, but from an Application Security perspective it does feel like a gap we're opening up.

The approach that adds an optional `dryRun: true` property to the top level `ChainCallDTO` makes it more clear that the end user is only authorizing a dry run of the method execution, not an actual submission. 

On the other hand, that means to use a dry run and then submit, an end user would need to sign two separate DTOs. This could be very inconvenient from a User Experience perspective...

Your approach only requires an end user sign once, which makes for a much smoother and more convenient flow than my approach. But end users still have to sign before dry runs can be executed. 

It might be worth soliciting discussion from some of our consumers / clients / implementers on the above points. I feel like an ideal scenario would let us execute dry runs without needing a DTO signed by the end user at all. Then they only need to sign if they want to go through with the actual submission.

dzikowski: I see your point. We should definitely not use signature in the inner dto to prevent from reply attacks. On the other hand we still need to have a way to authenticate the caller (we use signatures for that). Have a look at the fix: https://github.com/GalaChain/sdk/pull/281/commits/05aa2927cd0839c0cd31c24926f68023e1f46658

chaincode/src/contracts/GalaContract.ts
```diff
@@ -92,42 +102,89 @@ export abstract class GalaContract extends Contract {
...
+    // check if method exists
+    if (!methodName) {
+      const availableMethods = Array.from(methodNames.keys()).sort();
+      throw new NotFoundError(
+        `Method ${dto.method} not found. Available methods: ${availableMethods.join(", ")}`
+      );
+    }
+
+    // For dry run we don't use the regular authorization. We don't want users to provide signatures
+    // to avoid reply attack in case if the method is eventually not executed, and someone in the middle
```
sentientforest, line 163:
```suggestion
    // to avoid replay attack in case if the method is eventually not executed, and someone in the middle
```

sentientforest (approved):

PR #289, opened 2024-07-02
connect hotfix

befitsandpiper:
fix discrepancy with address private variable and setters getters, cleanup url references in derivative clients

### DIFF
--- a/chain-connect/src/GalachainConnectClient.ts
+++ b/chain-connect/src/GalachainConnectClient.ts
@@ -22,24 +22,24 @@ declare global {
 }
 
 export class GalachainConnectClient {
-  private address_: string;
-  private provider_: BrowserProvider | undefined;
-  private chainCodeUrl: string;
+  #address: string;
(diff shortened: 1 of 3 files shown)


chain-connect/src/GalachainConnectClient.ts
```diff
@@ -22,45 +22,45 @@ declare global {
...
+    this.#provider = new BrowserProvider(window.ethereum);
 
     try {
-      const accounts = (await this.provider_.send("eth_requestAccounts", [])) as string[];
-      this.address_ = accounts[0];
+      const accounts = (await this.#provider.send("eth_requestAccounts", [])) as string[];
+      this.address = accounts[0];
 
-      return this.address;
+      return this.#address;
```
erikgala, line 56: This should still just return this.address I think so the user is returned the valid GalaChain address

erikgala (approved):

RhiannonNichols (approved):

PR #293, opened 2024-07-09
Infinite allowances

befitsandpiper:
- allow non-mint allowances to be granted even if they exceed the balance a user has
- update logic to support infinite non-mint allowances (transfer, burn, use, etc)


Fuerback (approved):

chaincode/src/allowances/grantAllowance.spec.ts
```diff
@@ -109,10 +109,9 @@ describe("GrantAllowance", () => {
     expect(writes).toEqual({});
   });
 
-  it("should fail to GrantAllowance for fungible token when quantity is greater than balance", async () => {
+  it("should no longer fail to GrantAllowance for fungible token when quantity is greater than balance", async () => {
```
dzikowski, line 112:
```suggestion
  it("should allow to GrantAllowance for fungible token when quantity is greater than balance", async () => {
```

chaincode/src/allowances/grantAllowance.spec.ts
```diff
@@ -201,6 +197,50 @@ describe("GrantAllowance", () => {
...
+      created: ctx.txUnixTime,
+      quantity: new BigNumber(Infinity),
+      grantedBy: users.testUser2Id,
+      grantedTo: users.testUser1Id,
+      allowanceType: AllowanceType.Lock
+    }));
+    expect(response).toEqual(GalaChainResponse.Success([allowance]));
+    expect(writes).toEqual(writesMap(allowance));
+  });
+
```
dzikowski, line 243: I think there should be also tests that verify mint allowances:
- should fail to grant mint allowance for fungible token when quantity is greater than balance
- should fail to grant infinite mint allowance

chaincode/src/allowances/useAllowances.ts
```diff
@@ -98,8 +98,10 @@ export async function useAllowances(
 
     await putChainObject(ctx, newClaim);
 
-    // Update the allowance in the context
-    await putChainObject(ctx, tokenAllowance);
+    // Update the allowance in the context if quantity is finite
+    if (tokenAllowance.quantity.isFinite()) {
+      await putChainObject(ctx, tokenAllowance);
+    }
```
dzikowski, line 104: this is actually an inconsistency, in two ways:
1. We update `tokenAllowance` object before, and now we don't save changes. If you change some object that was read, while reading the code, you expect it to be saved, and the final if statement might be a surprise.
2. With no update on allowance, all `*Spent` fields remain at `0`, which also might be misleading. And I know we don't want to update it (performance). So we need to have a better default (`undefined`?). This way when infinite allowance is created it would get `undefined` for all `*Spent` fields and we won't update it at all.

chaincode/src/allowances/grantAllowance.spec.ts
befitsandpiper: The first one doesn't really make sense I don't think since mint allowances are granted by token authorities. So there isn't really a balance to compare against. But, the 2nd one is simple enough

dzikowski (approved):

PR #297, opened 2024-07-09
Dto updates

IndiaJonathan:


IndiaJonathan:
Draft WIP

Before I continue, do we want to go down this route in order to allow DTOs to properly not include some already defined fields?

dzikowski:
@jburk-gala I like your approach. The more type safety the better.

You might be also interested in the work we did here with `NonFunctionProperties<Dto>` type, it feels pretty similar: https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/dtos.ts#L64

we use it for instance in `createValidDto(DtoClass, { plain obj })`, where `plain obj` is actually verified by the compiler

the downside of this approach is that it requires plain objects that may differ from the ones that are produces by our serialization. For instance you need to provide `BigNumber` object, where the serialized value is a string

In your solution, I wonder if it is possible to add the default constructor on the parent class in a way that you don't need to provide it for your own. But maybe I want too much from the typescript compiler 😄

IndiaJonathan:
That is very similar, the reason I did this is because I don't want to strip out the functions except explicitly by name (in case it's an intentional inclusion)

On the parent, I don't think there's a way to do it to my knowledge after digging a bit

IndiaJonathan:
Moved from draft -> ready to merge

.gitignore
```diff
@@ -42,4 +42,5 @@ testem.log
 .DS_Store
 Thumbs.db
 .npmrc
-/.nx
\ No newline at end of file
+/.nx
+chain-cli/bin/run
```
dzikowski, line 46: I think we acually need this to make cli work

dzikowski (approved):

.gitignore
IndiaJonathan: Makes sense, removed that line!

PR #326, opened 2024-07-26
universal provider

befitsandpiper:
some initial work on TON client


chain-connect/src/GalachainConnectTONClient.ts
```diff
@@ -0,0 +1,184 @@
...
+        // const messageCell = beginCell()
+        // .storeBuffer(dtoBuff)
+        // .endCell();
+
+        // const signer = await this.#provider.getSigner();
+        // const signature = await signer.provider.send("personal_sign", [this.#tonAddress, dto]);
+
+        // const signature =
+
+        // return await this.submit(url, method, { ...prefixedPayload, signature }, headers);
```
dzikowski, line 133: yes, and here is the place where we need to get signature, and add it to DTO. Also we'll need to have an additional field in dto to determine the signing scheme.

This is the `enum` I'd like you to add to `dtos.ts`:

```typescript
export enum SigningScheme {
  ETH = "ETH",
  TON = "TON"
}
```

And this is the field to add to `ChainCallDTO` right after the `signerPublicKey`:

```typescript
  @JSONSchema({
    description:
      "Signing scheme used for the signature. ETH for Ethereum, TON for The Open Network." +
      `Default: ${SigningScheme.ETH}`
  })
  @IsOptional()
  @StringEnumProperty(SigningScheme)
  public signing?: SigningScheme;
```

I'm doing the same changes in #325, but I think you need it as well to implement the proper code without waiting till #325 is finished.

chain-connect/src/GalachainConnectTONClient.ts
```diff
@@ -0,0 +1,184 @@
...
+        console.error(tonProof.error);
+      }
+
+      if (!this.#accessToken) {
+        connector.disconnect();
+      }
+    });
+  }
+
+  private async checkProof(proof: TonProofItemReplySuccess["proof"], account: Account) {
```
dzikowski, line 73: I don't think you need to check proof in any way. You need to connect to TON and get the signature to add it to DTO. Proof/signature verification happens chaincode-side.

chain-connect/src/GalachainConnectTONClient.ts
dzikowski: Also one more thing for signing - I don't know how it looks from TON side, but signature verification may also have a seed/prefix, like `ton-connect` string here: https://github.com/ton-connect/demo-dapp-with-react-ui/blob/master/src/server/services/ton-proof-service.ts#L103

please add this (if needed) as a `prefix` in dto

PR #329, opened 2024-08-02
ETH V4 Signing

IndiaJonathan:
Adding in ETH V4 signing to the client by default, also supporting typed requests


dzikowski:
I have basically two comments/suggestions to this PR:

1. It seems the supporting EIP712 signing is mostly not a matter of the actual signing, but creating the payload to sign. So my first suggestion is to move the relevant part of the code  (and the `if` condition) to `getPayloadToSign` function. Please have a look at the attached git patch file - it applies my suggested changes while keeping tests green (also in connect library). [Move_EIP712_payload_creation_to_getPayloadToSign_function.patch](https://github.com/user-attachments/files/16494853/Move_EIP712_payload_creation_to_getPayloadToSign_function.patch)
2. My second comment is about ethers library. We were intentionally avoiding adding ethers dependency to chain-api library, because it is very heavy (about 12.6 MB of data), and we are using elliptic (119 kB). And we want to keep chain-api as lightweight as possible. Additionally, ethers depends on elliptic, and uses elliptic for signing. So adding ethers to the same project which has elliptic as a dependency is a risk of potential version conflicts and even more bloated node_modules. Do you think is there a way to avoid adding ethers library? Worst case - even copy the relevant code for `TypedDataEncoder.encode` from ethers?

chain-api/src/ethers/address.ts
```diff
@@ -0,0 +1,198 @@
+import { keccak256 } from "js-sha3";
+
+import { assertArgument } from "./errors";
+import { getBytes } from "./utils/data";
+
+const BN_0 = BigInt(0);
+const BN_36 = BigInt(36);
+
+function getChecksumAddress(address: string): string {
+  //    if (!isHexString(address, 20)) {
```
herr-bubu, line 10: is this comment needed?

chain-api/src/utils/signatures.ts
```diff
@@ -27,7 +28,33 @@ export class InvalidSignatureFormatError extends ValidationFailedError {}
...
+  value: EIP712Value;
+}
+
+// Type guard to check if an object is EIP712Object
+function isEIP712Object(obj: object): obj is EIP712Object {
+  return obj && typeof obj === "object" && "domain" in obj && "types" in obj;
+}
+
+function getEIP712PayloadToSign(obj: EIP712Object): string {
+  // return ethers.TypedDataEncoder.encode(obj.domain, obj.types, obj);
```
herr-bubu, line 48: is commented code needed here?

herr-bubu (approved):
🚀

chain-api/src/utils/signatures.ts
IndiaJonathan: Nope, removing it!

chain-api/src/ethers/address.ts
IndiaJonathan: Nope, removing i!

PR #344, opened 2024-08-27
Fix CURL command in trigger GitLab pipeline

dzikowski:
job failed with: https://github.com/GalaChain/sdk/actions/runs/10575975247/job/29301124852


RhiannonNichols (approved):
Looks like a small error on my part. Thank you for catching and fixing it! 

Approved

PR #347, opened 2024-08-29
BN serialization fix

IndiaJonathan:


chain-api/src/ethers/utils/utf8.ts
```diff
@@ -278,8 +280,15 @@ function getUtf8CodePoints(_bytes: BytesLike, onError?: Utf8ErrorFunc): Array<nu
...
-  assertArgument(typeof str === "string", "invalid string value", "str", str);
+export function toUtf8Bytes(str: string | BigNumber, form?: UnicodeNormalizationForm): Uint8Array {
+  assertArgument(
+    typeof str === "string" || BigNumber.isBigNumber(str),
+    `invalid string or bignumber value`,
+    "str or BN",
+    str
+  );
+
+  str = str.toString();
```
dzikowski, line 291: actually we use `bn.toFixed()` as a serialization for BigNumber objects. My recommendation is to do the same here, to keep it more readable (since toString produces scientific notation for bigger numbers)

chain-api/src/utils/serialize.spec.ts
```diff
@@ -122,3 +123,10 @@ it("should handle very large numbers with no decimals", () => {
   expect(chainObjectS).toEqual(expectedClassS);
   expect(plainS).toEqual(expectedPlainS);
 });
+
+it("Bignumber toUtf8Bytes", () => {
+  const bigNumber = BigNumber("300");
+  console.log(`Type: ${typeof bigNumber}`);
```
dzikowski, line 129: redundant console.log

chain-api/src/utils/serialize.spec.ts
```diff
@@ -122,3 +123,10 @@ it("should handle very large numbers with no decimals", () => {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  expect(toUtf8Bytes(bigNumber as any)).toEqual(toUtf8Bytes("300"));
```
dzikowski, line 131: I believe the type casting is no longer needed sice `toUtf8Bytes` accepts `string | BigNumber`

chain-api/src/ethers/utils/utf8.ts
IndiaJonathan: Changed it out! I did wrap it so as not to change the string if it's passed in. I can remove the check if that's better

chain-api/src/utils/serialize.spec.ts
IndiaJonathan: Removed!

chain-api/src/utils/serialize.spec.ts
IndiaJonathan: Yep, that was old. Stripping it out!

PR #348, opened 2024-08-29
Add CURATOR_ORG_MSP env support for allowedOrgs param in contracts

Ivan456:


chain-cli/chaincode-template/src/token/GalaChainTokenContract.ts
```diff
@@ -91,6 +91,8 @@ import { Info } from "fabric-contract-api";
 
 import { version } from "../../package.json";
 
+const curatorOrgMsp = process.env.CURATOR_ORG_MSP ?? curatorOrgMsp;
```
dzikowski, line 94: I believe that was the intention:

```suggestion
const curatorOrgMsp = process.env.CURATOR_ORG_MSP ?? "CuratorOrg";
```

chaincode/src/__test__/GalaChainTokenContract.ts
```diff
@@ -92,6 +92,9 @@ import {
 import { version } from "../../package.json";
 import { EVALUATE, SUBMIT } from "../contracts";
 
+const curatorOrgMsp = process.env.CURATOR_ORG_MSP ?? curatorOrgMsp;
+
```
dzikowski, line 96:
```suggestion
const curatorOrgMsp = process.env.CURATOR_ORG_MSP ?? "CuratorOrg";
```

dzikowski (approved):

PR #352, opened 2024-09-03
Dto updates

IndiaJonathan:
Minor fixes for types


chain-connect/src/GalachainClient.ts
```diff
@@ -129,10 +129,10 @@ export class GalachainClient implements CustomClient {
   }
 
   // PublicKey Chaincode calls:
-  public GetMyProfile(dto: GetMyProfileParams) {
-    return this.submit<UserProfileBody, GetMyProfileParams>({
+  public GetMyProfile() {
+    return this.submit<UserProfileBody, {}>({
       method: "GetMyProfile",
-      payload: dto,
+      payload: {},
```
befitsandpiper, line 135: why is the payload here being changed to an empty object?

befitsandpiper (approved):

PR #354, opened 2024-09-05
Alias Resolution

### LINKED ISSUE #247: Resolve user alias for asset methods (balances, transfer etc.)
Currently `UserProfile` is saved on `GCUP/<eth-addr>` key and contains the `alias` field. When you authorize with signature-based auth `ctx.callingUser` is set to `alias`. Also all balances and other user-related key parts use alias. But when you do some operations, like fetching balances or tranferring tokens, we don't search for aliases. If you provide `eth|<ath-addr>` for user that has an alias, it won't find the alias, which leads to potentially inconsistent state.

Implementation proposal:
1. For all DTO parameters which point to user, use a custom class `UserId` (like `owner?: UserId` in `FetchBalances`)
2. `UserId` should be serialized to string and deserialized from string to something like `{ "userId": string }`
3. For all ChainObject parameters which point to user, use a custom class - `UserAlias`.
4. `UserAlias` should be serialized to string containing user alias to keep backward compatibility (we use alias-first approach), and deserialized to something like `{ "alias": string }`.
5. Each time you want to use `UserAlias` type, you need to get it with some custom method like `getUserAlias(id: UserId): UserAlias`. This way we keep type-safety and enforce resolving user aliases.

To be considered:
1. Maybe we should implement it once we drop the legacy auth: #176

Jehosephat:
Resolves #247


dzikowski:
will be resolved in #411

PR #355, opened 2024-09-06
Galaconnect refactor

IndiaJonathan:
- Update gala connect to split out connections and actual apis so connections can be reused, this will also allow for custom contract wrapping later

Dependent on https://github.com/GalaChain/sdk/pull/352


chain-connect/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gala-chain/connect",
-  "version": "1.4.2",
+  "version": "1.4.3",
```
dzikowski, line 3: you don't need to provide this, next release version is `1.4.2`, and our CI creates automatic PRs with version bump after release, for instance: https://github.com/GalaChain/sdk/pull/350/files

```suggestion
  "version": "1.4.2",
```

dzikowski (approved):

PR #365, opened 2024-09-12
Initial migration of fees code from internal codebase to public sdk

sentientforest:
Migrate Fees code to public SDK
    
Includes all fee-related code and supporting oracle code.
Supporting types defined for bridging and swaps have been included,
but chaincode methods and other code for these features will
be migrated later.


chaincode/src/fees/galaFeeGate.spec.ts
```diff
@@ -0,0 +1,569 @@
...
+
+  const feeCodeDefinition2 = instanceToInstance(feeCodeDefinition1);
+  feeCodeDefinition2.feeThresholdUses = feeThreshold.plus(feeThreshold);
+  feeCodeDefinition2.baseQuantity = new BigNumber("100");
+
+  const feeCodeDefinition3 = instanceToInstance(feeCodeDefinition1);
+  feeCodeDefinition3.feeThresholdUses = new BigNumber(0);
+  (feeCodeDefinition3.feeAccelerationRateType = FeeAccelerationRateType.Multiplicative),
+    (feeCodeDefinition3.baseQuantity = new BigNumber("2"));
+  feeCodeDefinition3.feeAccelerationRate = new BigNumber("2");
```
dzikowski, line 62:
```suggestion
  feeCodeDefinition3.feeAccelerationRateType = FeeAccelerationRateType.Multiplicative;
  feeCodeDefinition3.baseQuantity = new BigNumber("2");
  feeCodeDefinition3.feeAccelerationRate = new BigNumber("2");
```

chaincode/src/fees/settleFeeBalances.ts
```diff
@@ -0,0 +1,83 @@
...
+  data: SettleFeeBalancesParams
+): Promise<SettleFeeBalancesResponse> {
+  const results = await getObjectsByKeys(ctx, FeePendingBalance, data.chainKeys);
+
+  const settlementResults: FeeBalanceSettlement[] = [];
+
+  const { year, month, day, hours, minutes, seconds } = txUnixTimeToDateIndexKeys(ctx.txUnixTime);
+  const txId = ctx.stub.getTxID();
+
+  for (const balance of results) {
```
dzikowski, line 43: can we do it in parallel with with `await Promise.all(results.map(async (balance) => { ... }));`?

I believe the only benefit we loose is random order of balances which may fail (if there are balance duplicates), but in that case the whole operation fails anyway

dzikowski: (or look at one of next comments and just ignore the above)

chaincode/src/fees/settleFeeCreditReceipts.ts
```diff
@@ -0,0 +1,42 @@
...
+  ctx: GalaChainContext,
+  data: SettleFeeCreditReceiptsParams
+): Promise<SettleFeeCreditReceiptsResponse> {
+  const results = await getObjectsByKeys(ctx, FeeBalanceCreditReceipt, data.chainKeys);
+
+  for (const receipt of results) {
+    receipt.status = FeeReceiptStatus.Settled;
+
+    await putChainObject(ctx, receipt);
+  }
```
dzikowski, line 35: I was going to suggest that we can do it in parallel as follows: 

```suggestion
  await Promise.all(results.map(async (receipt) => {
    receipt.status = FeeReceiptStatus.Settled;
    await putChainObject(ctx, receipt);
  }));
```

But from the other hand it is not saved immediately, but instead put to cache - which is somehow synchronous (all execution in the same thread). So ultimately it doesn't matter 🤔 And it won't speed up anything.

chaincode/src/fees/settleFeeBalances.ts
```diff
@@ -0,0 +1,83 @@
...
+      seconds,
+      creditToUser: balance.owner,
+      txId,
+      quantity: unspentBalance,
+      status: FeeReceiptStatus.Open
+    });
+
+    await receipt.validateOrReject();
+
+    await putChainObject(ctx, receipt);
```
dzikowski, line 71: for some time now `putChainObject` has this validation built in, no need to duplicate

chaincode/src/fees/splitFeeBurnAndTransfer.ts
```diff
@@ -0,0 +1,92 @@
...
+    await transferToken(ctx, {
+      from: payingUser,
+      to: transfer.transferToUser,
+      tokenInstanceKey: galaCurrencyKey,
+      quantity: transfer.transferQuantity,
+      allowancesToUse: [],
+      authorizedOnBehalf: undefined
+    }).catch((e) => {
+      throw new PaymentRequiredError(
+        `Failed to transferToken for splitFeeFormula: ` +
```
dzikowski, line 84: do we throw that error in all cases? don't we want to check the error type?

chaincode/src/fees/writeChannelPaymentReceipt.ts
```diff
@@ -0,0 +1,40 @@
...
+  status: FeeReceiptStatus;
+}
+export async function writeChannelPaymentReceipt(
+  ctx: GalaChainContext,
+  data: WriteChannelPaymentReceiptParams
+): Promise<void> {
+  const channelPaymentReceipt: FeeChannelPaymentReceipt = plainToInstance(FeeChannelPaymentReceipt, data);
+
+  await channelPaymentReceipt.validateOrReject();
+  await putChainObject(ctx, channelPaymentReceipt);
```
dzikowski, line 39:
```suggestion
  await putChainObject(ctx, channelPaymentReceipt);
```

dzikowski: (please double check that also in other places)

chaincode/src/oracle/saveOracleDefinition.ts
```diff
@@ -0,0 +1,64 @@
...
+    ctx,
+    OracleDefinition,
+    OracleDefinition.getCompositeKeyFromParts(OracleDefinition.INDEX_KEY, [data.name])
+  ).catch((e) => {
+    const chainError = ChainError.from(e);
+    if (chainError.matches(ErrorCode.NOT_FOUND)) {
+      return undefined;
+    } else {
+      throw chainError;
+    }
```
dzikowski, line 40: I believe this is a common pattern we have in our code, I wonder how we can improve our `ChainError` to address that. Maybe something like this would be useful?

```
  ).catch((e) => ChainError.ignore(e, ErrorCode.NOT_FOUND));
```

I think we should create an issue for that, or maybe some common other utils

dzikowski (approved):

chaincode/src/fees/splitFeeBurnAndTransfer.ts
sentientforest: I agree with the sentiment - however it's a little complicated in this case. 

From a cursory look, I imagine the `ValidationFailedError` in `TokenBalance` with the string message `Insufficient Balance` would be a primary case to check for and throw `PaymentRequired`. 

We could assume other errors occur for reasons unrelated to the user being unable to pay, and maybe `PaymentRequired` is inappropriate. In those cases, maybe we would throw the provided error as-is rather than creating a new `PaymentRequiredError`.

However it could be brittle to code for this exact error - there are lots of other `ValidationFailedError`s thrown in `TokenBalance` and `TransferToken` and it's not clear to me what would be the best way to code for that specific error (e.g. `e.code === ErrorCodes.VALIDATION_FAILED && e.message === "Insufficient Balance" would break if someone edits the message in `TokenBalance.ts`). We'd probably want to encode a new, specific error to cover the "insufficient balance" case. But it's also possible other errors would additionally make sense to be included (like ObjectNotFound when a balance is not found, because the user has no balance whatsoever).

And it may be useful for consuming clients to get the `paymentQuantity` value regardless, for display to the end user or other uses. And this value is only provided if we throw `PaymentRequiredError` and include it. Given that 

1) whatever error happens here, was caused as a result of attempting to handle a required payment
2) clients might need or want the payment quantity 
3) the `PaymentRequiredError` is including the message from anything thrown - so something unrelated to an insufficient balance could still bubble up here

(another reason I'm hesitant to rework this too much is I'm currently in-process porting this to the SDK while still maintaining / updating new feature requests in the version in our internal repository, double-duty until we get this merged and incorporated back).

So I would argue yes, in this case we always throw a `PaymentRequiredError` for the reasons above. Or we pursue it as an upgrade later in a separate issue, after we get the migration settled. 

However, one useful change I could make is to include the error code/type and optional payload from the thrown error as part of the `PaymentRequiredError` - it currently propagates up the message but not the other diagnostics.

chaincode/src/fees/settleFeeCreditReceipts.ts
sentientforest: Yeah, `Promise.all()` bit me once in the past, I forget exactly where it was in our assets code, but I had to switch it out to a loop. Ever since I find it a lot simpler to read and reason about when it's just in a loop like this.

PR #375, opened 2024-09-23
fix: prevent burnTokens() last-write-wins data integrity issues

sentientforest:
Patch the following two data integrity issues:

Data integrity issue 1: burnQuantities argument to burnTokens() contains two values with the same tokenInstanceKey properties, causing two debits to the user balance but only the last value's quantity being written to the `TokenBurn` chain object

Data integrity issue 2: compound chaincode that calls burnTokens() more than once within the same transaction will debit the TokenBalance correctly, but only the last call's quantity will reflect in the `TokenBurn` written to chain.


chaincode/src/burns/burnTokens.ts
```diff
@@ -209,3 +171,167 @@ export async function burnTokens(
...
+  }
+}
+
+/**
+ * If the `TokenBurnCounter` can be read, increment the quantity.
+ * Otherwise, create a new `TokenBurnCounter` with the quantity set to the provided value.
+ *
+ * @remarks
+ * The ChainKey design of `TokenBurnCounter` ranged objects ensures uniqueness by
+ * tokenInstanceKey properties, burnedBy, and a timeKey derived from the transaction timestamp.
```
dzikowski, line 271: technically, the transaction timestamp resolution is in seconds, which makes it unreliable if we want to use it for ensuring uniqueness. Don't you think it would be safer to use transaction ID instead? Is it a big deal?

dzikowski (approved):

chaincode/src/burns/burnTokens.ts
sentientforest: I agree its worth revisiting, however I did not want to increase the scope of this fix so I left as much as-is as I could. 

I imagine if we were to rework some chain keys we might also end up with a  data migration of some sort on our hands? And/or what downstream implications that could have on data consumers. 

Also, I think I looked into this before, and found that the `ctx.txUnixTime` is in fact giving us a timestamp in milliseconds. I could be mis-remembering. But if `fabric-node-sdk` is using JavaScript's `Date.now()` under the hood, it could explain why it returns a milliseconds timestamp. 

Milliseconds certainly add a bit more specificity, but I still agree that using a timestamp as the most specific chain key on any chain object (like `TokenBurn`) is always ill-advised and the transaction id is a better key for uniqueness.

PR #378, opened 2024-09-26
Connect documentation

IndiaJonathan:
Adding Gala connect documentation and examples


chain-connect/docs/Examples.md
```diff
@@ -0,0 +1,319 @@
...
+### Connecting to a Web3 Wallet
+
+First, you need to create a client and connect to a web3 wallet, the below shows a connection:
+
+```typescript
+import { BrowserConnectClient } from "@gala-chain/connect";
+import { ref } from "vue";
+
+const web3Wallet = new BrowserConnectClient();
+export const message = ref("");
```
erikgala, line 34: We should change this to be a generic typescript example, using vue is not necessary

chain-connect/docs/Examples.md
```diff
@@ -0,0 +1,319 @@
...
+  }
+}
+```
+
+### Template
+
+In your Vue component, you can add a simple template with a connect button and a button to create a token class:
+
+```html
+<template>
```
erikgala, line 108: This is something we could add in the UI package, but UI related examples are unnecessary here

chain-connect/docs/Examples.md
```diff
@@ -0,0 +1,319 @@
...
+  } catch (error) {
+    message.value = "Failed to generate wallet!";
+    console.error(error);
+  }
+}
+```
+
+### Template Integration
+
+In your Vue component, you can add a button to generate a wallet:
```
erikgala, line 222: Same thing here, remove examples related to html implementation using vue

chain-connect/src/customClients/TrustConnectClient.ts
```diff
@@ -91,72 +89,14 @@ function getTrustWalletFromWindow() {
   return window["trustwallet"] ?? null;
 }
 
-export class GalachainConnectTrustClient extends WebSigner {
+export class GalachainConnectTrustClient extends BrowserConnectClient {
```
erikgala, line 92: For naming consistency, both the file and class name here should be "TrustWalletConnectClient"

chain-connect/docs/Examples.md
```diff
@@ -0,0 +1,319 @@
...
+
+GalaChain Connect supports server-side signing using a private key. This is useful for backend services that need to interact with GalaChain without user interaction.
+
+### Server-Side Example
+
+```typescript
+import { PublicKeyApi, ServerSigningClient } from "@gala-chain/connect";
+
+const privateKey = "your-private-key-here";
+const connection = new ServerSigningClient(privateKey);
```
erikgala, line 254: I feel like we should change this class name to a more generic "SigningClient". This can be used both in browser and server sides, and using it in the browser client is a great example of creating a temporary wallet in games or dapps.

chain-connect/docs/Examples.md
IndiaJonathan: Changed!

chain-connect/docs/Examples.md
IndiaJonathan: Removed!

chain-connect/docs/Examples.md
IndiaJonathan: Removed!

chain-connect/docs/Examples.md
IndiaJonathan: Swapped the name out!

sentientforest (approved):

PR #379, opened 2024-09-27
Connect improvements

IndiaJonathan:


chain-connect/src/GalachainClient.ts
```diff
@@ -73,8 +78,12 @@ export abstract class CustomClient {
         return Promise.reject("Invalid JSON response");
       }
     }
-    return Promise.resolve(id ? { Hash: id, status: response.status } : { status: response.status });
+    throw new Error(`Unable to get data. Received response: ${JSON.stringify(response)}`);
```
IndiaJonathan, line 81: I'm not aware of any places that would need to return data without json, if anyone has one please let me know

befitsandpiper (approved):

PR #396, opened 2024-10-08
Gala chain connect gas fees

erikgala:
- Add chain gas fee contract methods to gala connect package
- Type fixes for ui packages and ignores pending method fixing
- New request/response types for fee objects


chain-connect/src/chainApis/GalaChainFeeApi.ts
```diff
@@ -0,0 +1,83 @@
...
+  ) {}
+
+  public AuthorizeFee(dto: FeeAuthorizationParams) {
+    return this.connection.submit<GalaChainResponse<FeeAuthorizationResponseBody>, FeeAuthorizationParams>({
+      method: "AuthorizeFee",
+      payload: dto,
+      sign: true,
+      url: this.chainCodeUrl
+    });
+  }
```
sentientforest, line 43: Is it possible to rework these methods to use DTOs rather than adding a `*Params` type for each? 

I would think that untyped data could be converted into a DTO class instance using `createValidDTO()`, and serialized using either the  `dto.serialize()` instance method or using `instanceToPlain` from the `class-validator` npm module to convert it into an object suitable for submitting to chain - although I haven't worked in the `connect` module much yet, I imagine I could be missing something.

sentientforest: This is a more detailed illustration of what I meant in my previous comment:

https://github.com/GalaChain/sdk/pull/400

Basically, working with the types that already exist, in a style that works similarly to how the `chain-api` module was built. 

The `@gala-chain/connect` library already depends on `@gala-chain/api`, so might as well also use `BigNumber`, `class-validator`, and `class-transformer`.

chain-api/src/types/TokenBalance.ts
```diff
@@ -150,21 +150,37 @@ export class TokenBalance extends ChainObject {
...
   @Type(() => TokenHold)
-  private inUseHolds?: Array<TokenHold>;
+  private _inUseHolds?: Array<TokenHold>;
 
   @BigNumberIsNotNegative()
   @BigNumberProperty()
-  private quantity: BigNumber;
+  private _quantity: BigNumber;
+
+  get instanceIds(){
```
erikgala, line 169: This is the attempt to make some of these vars public but readonly

sentientforest: I think changing `quantity` to `_quantity` might possibly modify the property name still, and possibly cause issues if it amounts to a schema change and we end up with a mismatch of existing `TokenBalance` objects using `quantity`, and new `TokenBalance` objects using `_quantity`. Would want to test that certainly. 

I wonder if changing `private` to `readonly` could serve your use case instead, while preserving the same write protection intended for these properties in chaincode?

https://www.typescriptlang.org/docs/handbook/2/classes.html#readonly

sentientforest:
@erikgala - make sure the existing modules will still build and test - I tried a fresh clone/download of your branch. Executing the following commands from the root of the mono repo leads to the following error -

```
$ npm install
$ npm run build

> @gala-chain/sdk@1.4.11 build
> nx run-many -t build


 NX   Could not load plugin @nx/storybook/plugin
```

chain-api/src/types/TokenBalance.ts
erikgala: Only problem with just setting to readonly is the properties are actually mutated in the code, which being readonly would prevent. Its like you need a private version to be read/write and a public version that is readonly

erikgala: Maybe I'm misunderstanding your comment about a mismatch of existing TokenBalance object, since the private properties are only accessible inside the class I'm not sure how that would be a problem.

erikgala: I've backed out those changes until a better fix can be implemented

sentientforest: > Maybe I'm misunderstanding your comment about a mismatch of existing TokenBalance object, since the private properties are only accessible inside the class I'm not sure how that would be a problem.

I haven't looked yet, but what I mean by this is:

If we looked in the block explorer, found some example `TokenBalance` objects written to chain, what would the names of the properties be and how might this change affect that? 

It would probably be whatever result we see when we serialize an instance of the class. I know properties like `quantity` are part of that serialized data model.

dzikowski (approved):

PR #400, opened 2024-10-08
Propose rework of chain-api and chain-connect type handling

sentientforest:
Removes need for duplication / redeclaration of chain-api classes for consuming libraries.
Standardize type transformation / validation across sub-projects of the mono repo
Use stronger / static typing

### DIFF
--- a/chain-connect/src/chainApis/TokenApi.ts
+++ b/chain-connect/src/chainApis/TokenApi.ts
@@ -13,45 +13,46 @@
  * limitations under the License.
  */
 import {
-  BatchMintTokenParams,
-  BurnTokensParams,
-  CreateTokenClassParams,
-  DeleteAllowancesParams,
+  BatchMintTokenDto,
+  BurnTokensDto,
+  CreateTokenClassDto,
+  DeleteAllowancesDto,
   FetchAllowancesBody,
-  FetchAllowancesParams,
-  FetchBalancesParams,
-  FetchBalancesWithPaginationParams,
+  FetchAllowancesDto,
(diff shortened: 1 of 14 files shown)


befitsandpiper (approved):
Looks good, I like the types simplification, thanks!

chain-connect/src/chainApis/TokenApi.ts
```diff
@@ -13,45 +13,46 @@
...
-  RefreshAllowanceParams,
-  ReleaseTokenParams,
+  MintTokenDto,
+  MintTokenWithAllowanceDto,
+  RefreshAllowanceDto,
+  ReleaseTokenDto,
   TokenAllowanceBody,
+  TokenBalance,
   TokenBalanceBody,
   TokenBurnBody,
```
erikgala, line 46: Are the body types still needed? I think those were also constructorArg wrapped types?

IndiaJonathan:
I think ideally we don't remove the types, allowing those makes devlopment much more intuitive. A typical dev is going to want to define the DTO with the type rather than always having to instantiate a class IMO

It also keeps it from being async

IndiaJonathan:
Basically, I think it's really important to keep the ability to do things like this:

```
  arbitraryToken: TokenClassKeyBody = {
    additionalKey: 'fakeTokenKey',
    category: 'fakeCategory',
    collection: 'fakeCollection',
    type: 'fakeType'
  }

  arbitraryTokenInstance: TokenInstanceKeyBody = { ...this.arbitraryToken, instance: '0' }
  
```

PR #403, opened 2024-10-09
Chain sale

befitsandpiper:
Sale chain object that allows a fulfiller to mint a new nft in exchange for a fungible token


chain-api/src/types/sale.ts
```diff
@@ -0,0 +1,503 @@
...
+export class TokenSaleDtoValidationError extends ValidationFailedError {
+  constructor(dtoName: string, errors: string[]) {
+    super(`${dtoName} validation failed: ${errors.join(". ")}`, {
+      dtoName,
+      errors
+    });
+  }
+}
+
+export class TokenSaleFailedError extends DefaultError {
```
dzikowski, line 499: `DefaultError` ends up with `500` http code, probably a different base - maybe validation failed also? It has `400`

chaincode/src/sales/createTokenSale.ts
```diff
@@ -0,0 +1,272 @@
...
+      const compositeKey = TokenClass.buildTokenClassCompositeKey(tokenClassKey);
+      const tokenClass = await getObjectByKey(ctx, TokenClass, compositeKey);
+
+      if(!tokenClass.authorities.includes(newSale.owner)) {
+        chainValidationErrors.push(
+          `${newSale.owner} is not an authority on ${tokenClassKey.toStringKey()}`
+        );
+      }
+
+      // TODO: Is this the correct way to check for remaining supply?
```
dzikowski, line 184: tbh I'm not sure - maybe @sentientforest ?

dzikowski (approved):
I assume this is not prod-ready, just a hackathon result to experiment on. Definitely requires testing and some cleanup before making it prod ready. Probably we can even rename the PR to something that points out that this is experimental. Like `Token sale experimental PoC` or something. This way it will be visible in release notes.

chaincode/src/sales/createTokenSale.ts
befitsandpiper: This has been fixed, and was backwards I believe

chain-api/src/types/sale.ts
befitsandpiper: yes, it is very experimental proof of concept. I think we do eventually want this in prod, but probably requires at least another week of work to be prod ready

PR #414, opened 2024-10-24
Connect bug fixes

erikgala:
- Add generic to object methods like GetObjectByKey and GetObjectHistory. Allows users to define expected response type.
- Fix dryrun method to stringify dto property before submitting
- Add pagination to FetchBalancesWithMetadata call, which is the correct request params


chain-connect/src/chainApis/GalaChainBaseApi.ts
```diff
@@ -23,19 +23,23 @@ export class GalaChainBaseApi {
     protected connection: GalaChainProvider
   ) {}
 
-  public DryRun(dto: DryRunRequest) {
+  public async DryRun(dto: DryRunRequest) {
+    await createValidDTO(DryRunDto, dto);
```
befitsandpiper, line 27: this looks extremely strange. We are creating a dto object, then stringifying it and duplicating the dto fields? I feel like this points to an interface failure, and something should change so we can just pass the dto around.

befitsandpiper (approved):
mostly looks good, a lil confused about the dto song and dance in the dryrun function

chain-connect/src/chainApis/GalaChainBaseApi.ts
erikgala: The problem is that the dto property from DryRunDto in chain-api is of type ChainCallDto. Whats expected in the API request though for the dto property is a string. This way, we can still validate the DTO is correct using createValidDto, but then its stringified so it is accepted in the API call. From what I could tell, the decorators on the dto property don't provide automatic conversion of the property from object to string when using instanceToPlain and plainToInstance. If this were the case, then we could avoid what I did here.

PR #418, opened 2024-10-29
Contructor args typings and error handling

erikgala:
Using NonFunctionParams was not recursively stripping functions in nested classes. This fixes that as well as fixes a bug with how errors were being handled.


jburk-gala (approved):
Awesome fixes!

IndiaJonathan (approved):

PR #432, opened 2024-11-12
Add command to get chaincode logs

umegbewe:


chain-cli/src/__mocks__/chalk.ts
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+const chalk = {
+  green: jest.fn((str: string) => str),
+  yellow: jest.fn((str: string) => str),
+  red: jest.fn((str: string) => str),
+  blue: jest.fn((str: string) => str),
+  gray: jest.fn((str: string) => str),
+  white: jest.fn((str: string) => str)
```
dzikowski, line 22: compilation fails here, I believe, to make it work, you can just provide for all functions:

```
green: (str: string) => str,
```

but why do you need to mock chalk?

dzikowski: oh, I see 🤔 

```
expect(chalk.green).toHaveBeenCalledWith("INFO ");
```

dzikowski: So I believe you can import Jest in this file:

```
import { jest } from '@jest/globals';
```

I believe we may have some misconfiguration in tsconfig files, because it's unavailable, and we do provide proper ` "compilerOptions": { "types": ["jest"] }` options

dzikowski (approved):

umegbewe:
CI passes, merging

PR #454, opened 2024-11-27
fix: post-mint-lock decimals handling v2

sentientforest:
CHAIN-819

fix for v2

### DIFF
--- a/chaincode/src/fees/exitGateImplementations.ts
+++ b/chaincode/src/fees/exitGateImplementations.ts
@@ -33,10 +33,10 @@ import {
 import BigNumber from "bignumber.js";
 
 import { lockTokens } from "../locks";
+import { resolveUserAlias } from "../services";
 import { GalaChainContext } from "../types";
 import { getObjectByKey } from "../utils";
 import { burnToMintProcessing } from "./extendedFeeGateProcessing";
-import { resolveUserAlias } from "../services";
 
 export interface IMintPostProcessing {
   tokenClass: TokenClassKey;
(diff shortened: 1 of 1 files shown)


chaincode/src/fees/exitGateImplementations.ts
```diff
@@ -68,9 +72,12 @@ export async function mintPostProcessing(ctx: GalaChainContext, data: IMintPostP
     return;
   }
 
+  const ownerAlias = await resolveUserAlias(ctx, owner);
```
dzikowski, line 75: 👍

chaincode/src/fees/exitGateImplementations.ts
```diff
@@ -86,7 +93,7 @@ export async function mintPostProcessing(ctx: GalaChainContext, data: IMintPostP
 export interface ILockOnMintProcessing {
   tokenClass: TokenClassKey;
   tokens: TokenInstanceKey[];
-  owner: UserAlias;
+  owner: string;
```
dzikowski, line 96: can it be `UserRef`? Typically we want to use `UserRef` for input, and `UserAlias` where we want the data to be unambiguous.

if you need, at some point transform a `string` to `UserRef`, you may use `asValidUserRef` function, which performs validation

dzikowski (approved):

chaincode/src/fees/exitGateImplementations.ts
sentientforest: This was my oversight, I think this change to `string` came through while I was reworking and getting more familiar with the `UserRef` vs `UserAlias` api. It's fixed now in a new commit.

PR #455, opened 2024-11-28
fix: lock authorities can unlock fungibles

sentientforest:
CHAIN-803

### DIFF
--- a/chain-api/src/types/TokenBalance.ts
+++ b/chain-api/src/types/TokenBalance.ts
@@ -467,7 +467,7 @@ export class TokenBalance extends ChainObject {
     return { lock };
   }
 
-  private isMatchingHold(
+  private isCallingUserAuthorized(
     hold: TokenHold,
     name?: string,
     callingUser?: string,
@@ -496,7 +496,7 @@ export class TokenBalance extends ChainObject {
 
     for (const hold of unexpiredLockedHolds) {
       // if neither the authority nor the name match, just leave this hold alone
-      if (!this.isMatchingHold(hold, name, callingUser, isTokenAuthority)) {
+      if (!this.isCallingUserAuthorized(hold, name, callingUser, isTokenAuthority)) {
         updated.push(hold);
         continue;
       }
--- a/chain-api/src/types/TokenBalance.ts
+++ b/chain-api/src/types/TokenBalance.ts
@@ -467,11 +467,18 @@ export class TokenBalance extends ChainObject {
     return { lock };
   }
 
-  private isMatchingHold(hold: TokenHold, name?: string, lockAuthority?: string): boolean {
+  private isMatchingHold(
+    hold: TokenHold,
+    name?: string,
+    callingUser?: string,
+    isTokenAuthority?: boolean
+  ): boolean {
     return (
-      (hold.name === name || (hold.name === undefined && name === undefined)) &&
-      (hold.lockAuthority === lockAuthority ||
-        (hold.lockAuthority === undefined && lockAuthority === undefined))
+      hold.name === name &&
+      (isTokenAuthority ||
+        callingUser === hold.lockAuthority ||
+        (hold.lockAuthority === undefined && callingUser === this.owner) ||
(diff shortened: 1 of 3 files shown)


chain-api/src/types/TokenBalance.ts
```diff
@@ -467,19 +467,27 @@ export class TokenBalance extends ChainObject {
...
+  ): boolean {
     return (
-      (hold.name === name || (hold.name === undefined && name === undefined)) &&
-      (hold.lockAuthority === lockAuthority ||
-        (hold.lockAuthority === undefined && lockAuthority === undefined))
+      hold.name === name &&
+      (isTokenAuthority ||
+        callingUser === hold.lockAuthority ||
+        (hold.lockAuthority === undefined && callingUser === this.owner) ||
+        (hold.lockAuthority === undefined && callingUser === hold.createdBy))
```
dzikowski, line 481: I was a bit hesitant to verify the token authority in balances, once I was doing the refactor, because I thought it was outside of the responsibility of the balance object. But seeing that changes, I think it makes sense, since we verify `lockAuthority` anyway. Token authority is just a special case of lock authority (it's always a token authority).

The only thing I think should be changed in that flow is that verifying authorities doesn't seem right for `isMatchingHold`. Ultimately I believe either:
1. `isMatchingHold` should just check the optional `name`, and the calling user and token owner/authority check should be inside `unlockToken`. This way we make the permission check not a responsibility of `TokenBalance`
2. or `isMatchingHold` should be renamed to something like `isCallingUserAuthorized` or `isAuthorizedToUnlock` (with calling user and also probably all other args required), so we are explicit it verifies the right to unlock. It also means the `name` arg becomes somewhat part of authorization (which makes sense imo)

Let me know what you think. We may also create a followup issue for that.

dzikowski (commented):
does it require a followup issue in SDK v2?

chain-api/src/types/TokenBalance.ts
sentientforest: With option 1) above, I see the following problem:

When you have a user that is:

1. Not the owner and 
2. Not a token authority

I think a refactor might get a little more complicated. 

If `isMatchingHold` only checks the `name` property, and we move authorization checks to `unlockToken` (and also other functions that call `ensureCanUnlockQuantity` i.e. `FillTokenSwap`) ... then we would need to either... 

a) match the `lockAuthority` for the above use case in `ensureCanUnlockQuantity`, and maybe add a separate bypass function for token authorities (currently there is no way to override the name/lock authority check in `ensureCanUnlockQuantity`
b) within `unlockToken` / `fillTokenSwap`, iterate through all the holds to find one in which the non-owner, non-authority user is the `lockAuthority`, and then extend the `TokenBalance` api to let us specify that exact hold (or multiple holds) or edit them directly ... 

Option 2) in your previous comment is more straightforward, and I agree, it would clarify the intent of the functionality. I'll change that within scope of the current issue. 

The logic here is a bit tricky and I'd like to make the smallest change possible to get the unlocking working, at least for v1 and the immediate priority. 

But here are the three primary use cases I see that are codified in the historical business logic, and maybe a v2 implementation or refactor makes sense as a new issue:

1. Owner can unlock any hold, as long as `lockAuthority` is undefined or equal to `owner`. `owner` cannot unlock tokens locked by a specified `lockAuthority`. 
2. Non-owner, non-token-authority can only unlock a hold for which they are the specified `lockAuthority`. 
3. A token authority can always unlock any hold, regardless if a `lockAuthority` is defined. Determining the list of token authorities requires reading the `TokenClass` from chain world state, and that action should definitely happen outside of the token balance. 

When token authorities unlock quantities off a balance, I also wonder if we want to consider which holds they unlock first, in the case of multiple token holds. Perhaps a token authority should prioritize unlocking holds with undefined `name` and/or `lockAuthority` properties first. 

I think it might make sense for the `TokenBalance` API to include a method specific for token authorities to use, which will act separately from the `owner`/`lockAuthority` logic. Probably something that should be pursued in a separate follow up issue.

dzikowski (approved):

PR #465, opened 2024-12-10
bump version to 1.5.10

sentientforest:
manually


sentientforest:
I'm not yet familiar enough with how our versioning Github actions are setup to know if they support auto-creating version bump commits for version 1.x now that we have released version 2.x. 

...guessing that `.github/workflows/manual-version-bump.yml` is the workflow to trigger. Looks like maybe the versions in `main-v1` and `main` are slightly different. 

However it's easy enough to run `node unifyVersions.js <new version number>` and create a pull request to `main-v1` manually, so I have done so here - it basically replicates what's in that workflow. 

This should get approved and merge prior to publishing the next 1.5.x version, to properly increment the versions across package.json files.

dzikowski (approved):

Issue #443, opened 2024-11-20
Token Sale Improvements


sentientforest:
Experimental / Prototype `TokenSale` functionality got merged previously in https://github.com/GalaChain/sdk/pull/403. 

This issue collections some feedback, improvements, and requested changes that didn't make it into the original pull request feedback before it got merged in. 

### 1) class definitions in `chain-api` should follow existing convention 

https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/sale.ts

Before we open-sourced `assets-api`, we used to have separate directories for `dtos` and `types`. The `types` directory contained classes that extended from `ChainObject` and the `dtos` directory contained classes that extended from `ChainCallDTO`. These were all mixed together in the same directory in the public sdk release. 

Currently all DTOs of a category are defined in a single file that is all lower-case (e.g. `swap.ts`) and each `ChainObject` is defined in a file named by Pascal case (aka upper Camel case), for example `TokenSwapRequest.ts`. 

I'm not defending the historical reasons that led to this prior convention, and I wasn't involved in the decision to combine these directories, but my recommendation would be to continue following established convention with any new feature like this until such time the convention is refactored out into some new style. 

Getting around to the actual ask: remove all classes that extend from `ChainObject` from `sale.ts`, and re-define them in new files named with the Pascal case convention. 

For example, `TokenSale` defined here: 

https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/sale.ts#L67

extends from `ChainObject`, so move it to a new file named `TokenSale.ts`. 

Whereas `CreateTokenSaleDto` extends from `ChainCallDTO`, so leave it in `sale.ts`:

https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/sale.ts#L280

### 2) Fix remaining `swap` --> `sale` renaming 

It looks like there are just a couple places where code copy/pasted from the `TokenSwap` features were not renamed with the `sale` semantic. I see a handful of cases in the `sale.ts` file in `chain-api`:

https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/sale.ts#L277
https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/sale.ts#L338
https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/sale.ts#L363
https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/sale.ts#L373
https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/sale.ts#L381
[...]

PR #468, opened 2024-12-12
migrate: swaps

sentientforest:
Migrate Token Swap features from assets-chaincode to public sdk

From the beginning, at least the start of my time on November 1, 2021, GalaChain has planned to support and included an implementation of peer-to-peer token
swapping.

This feature has been migrated, updated, and
refined through many refactorings and
changes in chaincode architecture. Many hands
contributed to this feature. It is only
now getting migrated to the public sdk for
wider use, due to various competing demands
throughout 2024. Finally, with
this commit, the remainder of the swap
implementation will be provided with the developer sdk.


sentientforest:
Related to https://github.com/GalaChain/sdk/issues/443, item 8)

Jehosephat (approved):

chaincode/src/swaps/fetchTokenSwapByRequestId.ts
```diff
@@ -0,0 +1,33 @@
...
+): Promise<TokenSwapRequest> {
+  let tokenSwap: TokenSwapRequest;
+
+  try {
+    tokenSwap = await getObjectByKey(ctx, TokenSwapRequest, swapRequestId);
+  } catch (error) {
+    throw new Error(`Token swap with swapRequestId ${swapRequestId} not found.`);
+  }
+
+  return tokenSwap;
```
dzikowski, line 32:
```suggestion
  return await getObjectByKey(ctx, TokenSwapRequest, swapRequestId);
```

why:
1. The error thrown indicates "not found" even if the underlying error in `getObjectByKey` is different
2. `Error` misses some useful properties that `ChainError` (thrown from `getObjectByKey`) has, like mapping to http 404.

I also don't think we need a custom error type for that function

chaincode/src/swaps/fetchTokenSwapsByInstance.ts
```diff
@@ -0,0 +1,150 @@
...
+    params.bookmark,
+    limit
+  );
+
+  const swapRequestIds = Array.from(
+    new Set(
+      instancesOffered.results.map((instanceOffered) => {
+        return instanceOffered.swapRequestId;
+      })
+    )
```
dzikowski, line 72:
```suggestion
  const swapRequestIds = Array.from(new Set(instancesOffered.results.map((i) => i.swapRequestId));
```

chaincode/src/swaps/fillTokenSwap.ts
```diff
@@ -0,0 +1,393 @@
...
+
+    // Transfer To Tokens - note that filledBy user doesn't have locked tokens
+    for (let index = 0; index < tokenSwap.wanted.length; index++) {
+      const toTokenId = tokenSwap.wanted[index].tokenInstance;
+      const currentQuantity = tokenSwap.wanted[index].quantity.times(uses);
+
+      if (filledOnBehalfOfOwner) {
+        // todo: unsure why verifyAndUseAllowances() requires *both* the tokenInstanceKey and tokenInstance be supplied -
+        // its code appears to just need the string properties not the actual TokenInstance off chain
+        // that might be worth reworking in the future
```
dzikowski, line 252: could you gather all of those problematic TODOs, and create a new issue for them? I believe it's a good way of enforcing more thorough review (by doing) of the business logic behind swaps

chaincode/src/swaps/requestTokenSwap.ts
```diff
@@ -0,0 +1,326 @@
...
+    tokenInstanceQty.tokenInstance = key;
+    tokenInstanceQty.quantity = t.quantity;
+    return tokenInstanceQty;
+  });
+
+  newSwap.txid = ctx.stub.getTxID();
+  newSwap.created = ctx.txUnixTime;
+  // 2023-11-16: added txid as new chain key because timestamp alone may not be precise enough to ensure uniqueness
+  // todo: determine if it would be straightforward to remove created as a ChainKey() from the TokenSwapRequest object
+  // i.e. if this feature has never been used in production, it could be easy
```
dzikowski, line 105: probably this comment needs some update

chaincode/src/swaps/ensureTokenSwapIndexing.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export async function ensureTokenSwapIndexing(
+  ctx,
+  params: EnsureTokenSwapIndexing
+): Promise<EnsureTokenSwapIndexingResponse> {
+  const results: TokenSwapRequest[] = await getObjectsByKeys(ctx, TokenSwapRequest, params.swapRequestIds);
+
+  let noOp = true;
+  const writes: ChainObject[] = [];
```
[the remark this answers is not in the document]
sentientforest: Technically yes, it appears that way. 

However I'm not sure if GalaSwap/Connect or other consumers are relying on this boolean being present in the response, so I would be hesitant to change it without looking into that. 

The scope of the initial task was just to migrate code as-is, so I've left this as I found it.

chaincode/src/swaps/fillTokenSwap.ts
sentientforest: Agree, I have written up a new issue here:

https://github.com/GalaChain/sdk/issues/470

It summarizes both high level goals of a review/refactor given new features introduced in the SDK since it was originally written, and it summarizes your comments on this pull request into a bulleted list.

Issue #446, opened 2024-11-20
Lock Authority cannot Unlock Tokens they do not own


sentientforest:
When defined, a `lockAuthority` on a `TokenHold` is expected to be able to unlock the token(s) regardless of ownership:

https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/TokenBalance.ts#L583-L591

However, the `unlockFungibleTokens` method does not support this logic:

https://github.com/GalaChain/sdk/blob/main/chaincode/src/locks/unlockToken.ts#L91-L105
https://github.com/GalaChain/sdk/blob/main/chain-api/src/types/TokenBalance.ts#L583-L591

Based on a read of the logic above, the only scenarios that will result in a successful unlock are:

1. calling user is the owner AND the owner is set as the `lockAuthority` on the `TokenHold`
2. calling user is a Token Authority AND the owner is set as the `lockAuthority` on the `TokenHold`

This issue tracks the following fixes:

1. owner can unlock fungible tokens when `lockAuthority` is undefined on the `TokenHold`
2. lock authorities can unlock fungible tokens they don't own, regardless of whether or not they are a token authority 
3. token authorities can unlock fungible tokens regardless of whether or not a lock authority is defined

PR #469, opened 2024-12-12
fix: lock authorities and token authorities can unlock fungibles v2

sentientforest:
Port changes from
https://github.com/GalaChain/sdk/pull/455 to v2

Closes https://github.com/GalaChain/sdk/issues/446

When defined, a lockAuthority on a TokenHold can unlock the token regardless of ownership.

A token authority can unlock a TokenHold on another users balance regardless of lockAuthority status.

Refactoring or reworking order of unlocks by token authorities is out of scope for this immediate fix, should be pursued in a separate issue if needed.


chain-api/src/types/TokenBalance.ts
```diff
@@ -425,19 +425,27 @@ export class TokenBalance extends ChainObject {
     this.lockedHolds = [...this.getUnexpiredLockedHolds(hold.created), hold];
   }
 
-  private isMatchingHold(hold: TokenHold, name?: string, lockAuthority?: string): boolean {
+  private isCallingUserAuthorized(
+    hold: TokenHold,
+    name?: string,
+    callingUser?: string,
```
dzikowski, line 431: I think we can make it safer at type level by:
```suggestion
    callingUser?: UserAlias,
```

this way we can enforce on developers to make proper resolution of user alias before. User alias is unambiguous for the user, while string or user ref may be ambiguous.

chain-api/src/types/TokenBalance.ts
```diff
@@ -425,19 +425,27 @@ export class TokenBalance extends ChainObject {
...
+        (hold.lockAuthority === undefined && callingUser === hold.createdBy))
     );
   }
 
   public unlockQuantity(
     quantity: BigNumber,
     currentTime: number,
     name?: string,
-    lockAuthority?: string
+    callingUser?: string,
```
dzikowski, line 447:
```suggestion
    callingUser?: UserAlias,
```

dzikowski (approved):

PR #475, opened 2024-12-19
Allow access for non-registered users

dzikowski:


chaincode/src/contracts/PublicKeyContract.ts
```diff
@@ -67,18 +65,7 @@ export class PublicKeyContract extends GalaContract {
...
-
-    if (profile === undefined) {
-      throw new NotFoundError(`UserProfile not found for ${ctx.callingUser}`, {
-        signature: dto.signature,
-        user: ctx.callingUser
-      });
-    }
-
-    return profile;
+    return ctx.callingUserProfile;
```
dzikowski, line 68: this way we can also return default profile here

chaincode/src/contracts/authenticate.ts
```diff
@@ -138,7 +138,11 @@ async function getUserProfile(
   const profile = await PublicKeyService.getUserProfile(ctx, address);
 
   if (profile === undefined) {
-    throw new UserNotRegisteredError(address);
+    if (ctx.config.allowNonRegisteredUsers) {
+      return PublicKeyService.getDefaultUserProfile(publicKey, signing);
+    } else {
+      throw new UserNotRegisteredError(address);
+    }
   }
```
dzikowski, line 146: this it the location where we make use of the feature flag

ggtravis (approved):

PR #481, opened 2025-01-20
Align recently added features with SDK v2 conventions

dzikowski:


chain-api/src/types/mint.ts
```diff
@@ -484,3 +484,17 @@ export class FetchTokenMintConfigurationsResponse extends ChainCallDTO {
   @IsString()
   bookmark: string;
 }
+
+export class DeleteTokenMintConfigurationDto extends SubmitCallDTO {
```
dzikowski, line 488: It is required to provide some additional dtos, since the default used (like `TokenClassKey` does not have `uniqueKey` defined)

Jehosephat (approved):

PR #484, opened 2025-01-29
Vesting Token

Jehosephat:


chain-api/src/types/Allocation.ts
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import BigNumber from "bignumber.js";
+import { IsInt, IsString, Min } from "class-validator";
+
+import { BigNumberIsNotNegative, BigNumberProperty, IsUserAlias } from "../validators";
+
+export class Allocation {
```
dzikowski, line 20: It would be good also to have the documentation of those fields (it will be visible in swagger). Also fields like `vestingPeriod` may include the unit in names (like: `vestingDays` or `vestingSeconds`)

chaincode/src/vesting/vestingToken.spec.ts
```diff
@@ -0,0 +1,558 @@
...
+      start: 123,
+      vestingName: "SuperTokenTGE",
+      //vestingTimeIncrement: "Days",
+      allocations: [
+        {
+          name: "allocation1",
+          owner: users.testUser1Id,
+          quantity: new BigNumber(100),
+          cliff: new BigNumber(90),
+          vestingPeriod: new BigNumber(730)
```
dzikowski, line 50: regular numbers are used for allocations

chaincode/src/vesting/vestingToken.spec.ts
```diff
@@ -0,0 +1,558 @@
...
+    const { ctx, contract, writes } = fixture(GalaChainTokenContract)
+      .callingUser(users.testUser1Id)
+      .savedState(currency.tokenClass()); //,balances?);
+
+    // When
+    const response = await contract.TransferToken(ctx, vestingTokenDto);
+
+    // Then
+    expect(response).toEqual(
+      GalaChainResponse.Success()
```
dzikowski, line 83: typically we use `transactionSuccess()` from `@gala-chain/test` library as success matcher

chain-api/src/types/VestingToken.ts
```diff
@@ -0,0 +1,54 @@
...
+  @ChainKey({ position: 2 })
+  @IsNotEmpty()
+  public type: string;
+
+  @ChainKey({ position: 3 })
+  @IsDefined()
+  public additionalKey: string;
+
+  @IsString()
+  vestingName: string;
```
dzikowski, line 43: if we are going to support multiple vesting names for a single token, we may consider adding it as a key part

chain-api/src/types/Allocation.ts
Jehosephat: Yeah, I was wrestling with whether to have a separate field for a enum with different period units. But since I'm pretty sure we're going with days I'll update that. Thanks!

chain-api/src/types/VestingToken.ts
Jehosephat: I don't think so. I considered not even having this field and just using the class key as the first part of the lock name.

chaincode/src/vesting/vestingToken.spec.ts
Jehosephat: Yeah, have quite a bit to do yet on the tests.

chaincode/src/vesting/vestingToken.ts
```diff
@@ -0,0 +1,174 @@
...
+            type: vestingToken.type,
+            additionalKey: vestingToken.additionalKey,
+            owner: allocation.owner
+        })
+
+        if(allocationBalance.length !== 1){
+            // throw some kind of error, this is weird
+        }
+
+        vestingTokenInfo.allocationBalances.push(allocationBalance[0])
```
sentientforest, line 85: This could be reworked to use `getObjectByKey` instead of `fetchBalances`. The token class key properties plus the owner represent all `ChainKey` properties defined on the `TokenBalance`.

chaincode/src/vesting/vestingToken.ts
```diff
@@ -0,0 +1,174 @@
...
+    // allocations add up to total supply/max cap
+
+    const tokenClassParams: CreateTokenClassParams = {
+        ...params
+    }
+    const tokenClassResponse = await createTokenClass(ctx, tokenClassParams)
+
+    const tokenInstanceKey = plainToInstance(TokenInstanceKey, {
+        ...params.tokenClass,
+        instance: "0"
```
sentientforest, line 105:
```suggestion
        instance: TokenInstance.FUNGIBLE_TOKEN_INSTANCE
```

chaincode/src/vesting/vestingToken.ts
```diff
@@ -0,0 +1,174 @@
...
+
+            const lockResponse = await lockToken(ctx, {
+                owner: allocation.owner,
+                lockAuthority: ctx.callingUser,
+                tokenInstanceKey,
+                quantity: currentLockQuantity,
+                allowancesToUse: [],
+                name: `${params.vestingName}-${allocation.name}-${i}`,
+                expires: expiration + daysToMilliseconds(i),
+                verifyAuthorizedOnBehalf: async () => undefined
```
sentientforest, line 150:
```suggestion
                verifyAuthorizedOnBehalf: async () => { callingOnBehalf: allocation.owner }
```

chain-api/src/types/VestingToken.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  @Min(0)
+  @IsInt()
+  startDate: number;
+
+  @ValidateNested({ each: true })
+  @Type(() => Allocation)
+  @ArrayMinSize(1)
+  @ArrayUnique()
+  allocations: Array<Allocation>;
```
sentientforest, line 53: Another modeling to consider

Rather than a `VestingToken` chain entry that references an array of `Allocation` objects, make a single, less nested model that just defines everything on a `VestingAllocation` object. 

Combine the chain keys, vestingName, startDate, name, owner, quantity, cliff, vestingPeriod on to separate chain entries. There would be a need for `Allocation.name` and/or `Allocation.owner` to be an additional chain key. 

Then the chaincode would use `getObjectsByPartialCompositeKey` rather than pulling the `VestingToken` and then iterating its allocations array. 

It would be a flatter and simpler structure, possibly yielding better error messaging or handling compared to nested validation errors. The tradeoff might be duplication of some data, like token collection, category, etc. properties.

Jehosephat: Thought about this a bit, definitely has it's merits. I stuck with the original model mostly because I think that folks looking at the chaincode objects will ultimately have an easier time understanding the associations this way.

sentientforest (approved):

PR #517, opened 2025-03-07
Add feature dex protocol fees and updated FeeGateCodes

Divyansh-Gupta-Antier:
Added dex protocol fees and updated a few calculations. Added dex and launchpad functions to FeeGateCodes.


chain-api/src/types/DexDtos.ts
```diff
@@ -660,7 +656,6 @@ export class GetRemoveLiqEstimationResDto extends ChainCallDTO {
 export class CollectProtocolFeesResDto extends ChainCallDTO {
   @IsNotEmpty()
   @BigNumberProperty()
```
sentientforest, line 658:
```suggestion

  @BigNumberProperty()
```

I don't recommend combining `@IsNotEmpty` with BigNumber related decorators

chain-api/src/types/DexDtos.ts
```diff
@@ -675,9 +670,19 @@ export class CollectProtocolFeesResDto extends ChainCallDTO {
 export class SetProtocolFeeResDto extends ChainCallDTO {
   @IsNumber()
   @IsNotEmpty()
```
sentientforest, line 672:
```suggestion
  @IsNumber()

```

If this number should not be negative, considering including `@IsPositive()`.

chain-api/src/types/DexDtos.ts
```diff
@@ -675,9 +670,19 @@ export class CollectProtocolFeesResDto extends ChainCallDTO {
...
   }
 }
+
+export class ConfigureDexFeeAddressDto extends ChainCallDTO {
+  @IsOptional()
+  @IsString()
+  public newDexFeeAddress?: string;
+
+  @IsOptional()
+  @Type(() => String)
```
sentientforest, line 686:
```suggestion
  @ValidateNested({ each: true })
  @Type(() => String)
```

chain-api/src/types/DexFeeConfig.ts
```diff
@@ -0,0 +1,46 @@
...
+import { ArrayNotEmpty, IsNumber, IsString, Max, Min } from "class-validator";
+
+import { ChainObject } from "./ChainObject";
+
+export class DexFeeConfig extends ChainObject {
+  @Exclude()
+  public static INDEX_KEY = "GCDPFC"; // GalaChain Dex Protocol Fee Configuration
+
+  @ArrayNotEmpty()
+  @IsString({ each: true })
```
sentientforest, line 25:
```suggestion
  @ValidateNested({ each: true })
  @Type(() => String)
```

chain-api/src/types/DexV3Pool.ts
```diff
@@ -356,12 +366,13 @@ export class Pool extends ChainObject {
...
+        zeroForOne,
+        state.sqrtPrice
       );
 
-      //cap the tick in valid range i.e. -887272 < tick < 887272
-      if (step.tickNext < -887272) throw new Error("Pool lacks enough liquidity to perform this operation");
-      if (step.tickNext > 887272) throw new Error("Pool lacks enough liquidity to perform this operation");
+      //cap the tick in valid range i.e. MIN_TICK < tick < MAX_TICK
+      if (step.tickNext < Pool.MIN_TICK || step.tickNext > Pool.MAX_TICK)
+        throw new Error("Not enough liquidity available in pool");
```
sentientforest, line 375:
```suggestion
      if (step.tickNext < Pool.MIN_TICK || step.tickNext > Pool.MAX_TICK) {
        throw new Error("Not enough liquidity available in pool");
      }
```

For an if-statement this long, prefer braces over one-line syntax. The existing code looks like one-line syntax spread across two lines, which could lead to confusion or mis-interpretation in the future.

chain-api/src/utils/dex/tick.helper.ts
```diff
@@ -132,6 +135,16 @@ export function flipTick(bitmap: Bitmap, tick: number, tickSpacing: number) {
...
 }
 
+function isTickInitialized(tick: number, tickSpacing: number, bitmap: Bitmap): boolean {
+  tick /= tickSpacing;
+  const [word, pos] = position(tick);
+  const mask = BigInt(1) << BigInt(pos);
+  if (bitmap[word] === undefined) return false;
+  const currentMask = BigInt(bitmap[word]);
+  const newMask = currentMask ^ mask;
+  return newMask == BigInt(0);
```
sentientforest, line 145:
```suggestion
  const mask = BigInt(1) << BigInt(pos);

  if (bitmap[word] === undefined) return false;

  const currentMask = BigInt(bitmap[word]);
  const newMask = currentMask ^ mask;

  return newMask == BigInt(0);
```

Newlines improve readability around the one-line if statement.

chain-cli/chaincode-template/e2e/dex-api.spec.ts
```diff
@@ -302,6 +368,54 @@ describe("DEx v3 Testing", () => {
...
+        token0,
+        token1,
+        token0Slipped,
+        token1Slipped
+      ).signed(user.privateKey);
+
+      const addLiquidityRes = await client.dexV3Contract.addLiquidity(dto);
+      expect(addLiquidityRes.Message).toBe("liquidity crossed max liquidity");
+    });
+    test("Adding liquidity equal to zero will throw error", async () => {
```
sentientforest, line 395:
```suggestion
    });

    test("Adding liquidity equal to zero will throw error", async () => {
```

Cleaner

chain-cli/chaincode-template/e2e/dex-api.spec.ts
```diff
@@ -302,6 +368,54 @@ describe("DEx v3 Testing", () => {
...
+        tb,
+        token0,
+        token1,
+        token0Slipped,
+        token1Slipped
+      ).signed(user.privateKey);
+
+      const addLiquidityRes = await client.dexV3Contract.addLiquidity(dto);
+      expect(addLiquidityRes.Message).toBe("Invalid Liquidity");
+    });
```
sentientforest, line 418:
```suggestion
    });

```

Improved readability

chain-cli/chaincode-template/e2e/dex-api.spec.ts
```diff
@@ -484,6 +598,15 @@ describe("DEx v3 Testing", () => {
...
     });
+    test("should throw error in estimating swap while swap more than avaiable liquiidity", async () => {
+      const amountToSwap = new BigNumber("200000000000000000000000000"),
+        sqrtPriceLimit = new BigNumber(1);
+      const dto = new SwapDto(ETH_ClassKey, USDT_ClassKey, fee, amountToSwap, true, sqrtPriceLimit).signed(
+        user.privateKey
+      );
+      const expectSwapRes = await client.dexV3Contract.quoteExactAmount(dto);
+      expect(expectSwapRes.Message).toBe("Not enough liquidity available in pool");
+    });
```
sentientforest, line 609:
```suggestion
    });

```

chain-cli/chaincode-template/e2e/dex-api.spec.ts
```diff
@@ -672,6 +802,25 @@ describe("DEx v3 Testing", () => {
...
+        USDT_ClassKey,
+        fee,
+        new BigNumber("92271.497628802094407217"),
+        ta,
+        tb,
+        user.identityKey
+      );
+      const removeLiqEstimation = await client.dexV3Contract.burnEstimate(dto);
+      expect(removeLiqEstimation.Message).toBe("Uint Out of Bounds error :Uint");
+    });
```
sentientforest, line 823:
```suggestion
    });

```

chain-cli/chaincode-template/e2e/dex-api.spec.ts
```diff
@@ -711,7 +860,26 @@ describe("DEx v3 Testing", () => {
         tb
       ).signed(user.privateKey);
       const collectRes = await client.dexV3Contract.collect(dto);
+      expect(collectRes.Message).toBe("Less balance accumulated");
+      const getPositionsDto = new GetUserPositionsDto(user.identityKey).signed(user.privateKey);
+      await client.dexV3Contract.getUserPositions(getPositionsDto);
+    });
```
sentientforest, line 866:
```suggestion
    });

```

chain-cli/chaincode-template/e2e/dex-api.spec.ts
```diff
@@ -1739,6 +2008,50 @@ describe("DEx v3 Testing", () => {
...
+        const configPlatformFeeAddressDTO = new ConfigureDexFeeAddressDto();
+        configPlatformFeeAddressDTO.newAuthorities = [];
+
+        configPlatformFeeAddressDTO.sign(authorityUser.privateKey);
+        const configRes = await client.dexV3Contract.configureDexFeeAddress(configPlatformFeeAddressDTO);
+
+        expect(configRes.Status).toEqual(0);
+        expect(configRes.Message).toEqual("At least one user should be defined to provide access");
+      });
+      test("It will revert if none of the input field are present", async () => {
```
sentientforest, line 2022:
```suggestion
      });

      test("It will revert if none of the input field are present", async () => {
```

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
```diff
@@ -1146,7 +1125,7 @@ describe("LaunchpadContract", () => {
       );
     });
 
-    test("SellWithExactTokens || It should revert if Expected Gala tokens from this operation exceeds the actual amount that will be provided.", async () => {
+    test("SellWithExactTokens || It should rever if Expected Gala tokens from this operation exceeds the actual amount that will be provided.", async () => {
```
sentientforest, line 1128:
```suggestion
    test("SellWithExactTokens || It should revert if Expected Gala tokens from this operation exceeds the actual amount that will be provided.", async () => {
```

Unintentional typo introduced?

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
```diff
@@ -1175,7 +1154,7 @@ describe("LaunchpadContract", () => {
 
       sellWithExactDTO.sign(user1.privateKey);
 
-      const sellWithNativerRes = await client.Launchpad.SellToken(sellWithExactDTO);
+      const sellWithNativerRes = await client.Launchpad.SellExactToken(sellWithExactDTO);
```
sentientforest, line 1157:
```suggestion
      const sellWithNativeRes = await client.Launchpad.SellExactToken(sellWithExactDTO);
```

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
```diff
@@ -1175,7 +1154,7 @@ describe("LaunchpadContract", () => {
...
 
       expect(sellWithNativerRes.Message).toEqual(
```
sentientforest, line 1159:
```suggestion
      expect(sellWithNativeRes.Message).toEqual(
```

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
```diff
@@ -1206,36 +1185,36 @@ describe("LaunchpadContract", () => {
       expect(fetchSaleRes.Data?.saleStatus).toEqual(SaleStatus.ONGOING);
     });
     test("It will revert if none of the input field are present", async () => {
```
sentientforest, line 1187:
```suggestion
    });

    test("It will revert if none of the input field are present", async () => {
```

sentientforest:
You are missing the code that ties together your new [Fee Gates](chaincode/src/fees/feeGateImplementations.ts) / [Fee Codes](chain-api/src/types/FeeGateCodes.ts) with their corresponding contract methods. 

The `@GalaTransaction` decorator includes options to configure logic that should execute `before` or `after` the main transaction logic, and only succeed atomically with all other writes. Ref: https://github.com/GalaChain/sdk/blob/main-v1/chaincode/src/contracts/GalaTransaction.ts#L172-L174

GalaChain Fees, the ecosystem Transaction Fees" (aka gas) are executed using these `before` hooks. The public SDK still does not seem to include the `GalaTokenContract` source code with fees as a reference, and this feature is not yet well documented. 

Basically though, given your new contracts, you would want this:

https://github.com/GalaChain/sdk/blob/main-v1/chaincode/src/contracts/DexV3Contract.ts#L74-L80

```
  @Submit({
    in: CreatePoolDto,
    out: Pool
  })
  public async CreatePool(ctx: GalaChainContext, dto: CreatePoolDto): Promise<Pool> {
    return await createPool(ctx, dto);
  }
```

To become this:

```
  @Submit({
    in: CreatePoolDto,
    out: Pool,
    before: createPoolFeeGate
  })
  public async CreatePool(ctx: GalaChainContext, dto: CreatePoolDto): Promise<Pool> {
    return await createPool(ctx, dto);
  }
```

Repeat for any contract methods that can potentially be fee-gated. 

Of course you also need to import your functions defined in `feeGateImplementations.ts` to their respective contracts, in order to then pass them to the `before` hook of the appropriate contract methods. 

```
import { createPoolFeeGate, /* etc */ } from "../fees";
```

As an aside, it's not strictly necessary to use different functions for different contract methods here, that's just what was done historically for maximum flexibility.

chaincode/src/dex/setProtocolFee.ts
```diff
@@ -36,20 +36,46 @@ export async function setProtocolFee(
...
+export async function configureDexFeeAddress(ctx: GalaChainContext, dto: ConfigureDexFeeAddressDto) {
+  if (!dto.newAuthorities?.length) {
+    throw new Error("At least one user should be defined to provide access");
+  }
+
+  const curatorOrgMsp = process.env.CURATOR_ORG_MSP ?? "CuratorOrg";
+
+  if (ctx.clientIdentity.getMSPID() !== curatorOrgMsp) {
+    throw new UnauthorizedError(`CallingUser ${ctx.callingUser} is not authorized to create or update`);
+  }
```
sentientforest, line 68: The current logic will prevent any non-curator user from executing this method, even if they are an authority on an existing `DexFeeConfig` entry on-chain. The code following this block implies you do want to permit authorities to update these configs. 

Consider moving the org check after the lookup from on-chain world state, similar to the `saveOracleDefinition` method:

https://github.com/GalaChain/sdk/blob/main-v1/chaincode/src/oracle/saveOracleDefinition.ts#L45-L47

```
  if (ctx.clientIdentity.getMSPID() !== curatorOrgMsp) {
    if (!existingOracle || !existingOracle.authorities.includes(ctx.callingUser)) {
```

Essentially, 

if you're not a Curator

and 

The oracle does not yet exist, or it does exist but you're not an authority, 

then throw this error.  This approach serves a middle ground where some authority is needed to get new configs on chain, preventing random or malicious users arbitrarily creating these entries, while giving less-permissioned users the ability to modify at will.

chaincode/src/fees/feeGateImplementations.ts
```diff
@@ -511,3 +511,43 @@ export async function combinedMintFees(ctx: GalaChainContext, data: ICombinedMin
...
+  return galaFeeGate(ctx, { feeCode: FeeGateCodes.SellExactToken });
+}
+
+export async function buyWithNativeFeeGate(ctx: GalaChainContext, dto: ChainCallDTO) {
+  return galaFeeGate(ctx, { feeCode: FeeGateCodes.BuyWithNative });
+}
+
+export async function sellWithNativeFeeGate(ctx: GalaChainContext, dto: ChainCallDTO) {
+  return galaFeeGate(ctx, { feeCode: FeeGateCodes.SellWithNative });
+}
```
sentientforest, line 553: Refer to my related top-level comment on fees: https://github.com/GalaChain/sdk/pull/517#issuecomment-2707320044

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
```diff
@@ -186,15 +165,15 @@ describe("LaunchpadContract", () => {
...
-    if (balaneData === undefined || balaneData.length === 0) {
+    const balanceData = fetchBalanceRes.Data;
+    if (balanceData === undefined || balanceData.length === 0) {
       return new BigNumber(0);
     }
 
-    balaneData[0].getQuantityTotal = function (): BigNumber {
+    balanceData[0].getQuantityTotal = function (): BigNumber {
       return this.quantity;
     };
```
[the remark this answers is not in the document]
Divyansh-Gupta-Antier: I am unable to use the `getQuantityTotal` function from `GalaChainResponse.Data` at runtime, as it results in an error stating that the function does not exist. Additionally, I cannot instantiate a new `TokenBalance` object because the `quantity` parameter is still treated as a private value, preventing direct access without a getter function. To work around this, I reimplemented the getter function for the object.

chain-api/src/types/DexFeeConfig.ts
Divyansh-Gupta-Antier: The `@ValidateNested` decorator isn't working properly with the string primitive data type. We have opted to use `@IsString({each: true})` instead.

chain-api/src/types/DexDtos.ts
Divyansh-Gupta-Antier: The `@ValidateNested` decorator isn't working properly with the string primitive data type. We have opted to use `@IsString({each: true})` instead.

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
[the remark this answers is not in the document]
sentientforest: So you tried the constructor approach, but not the `plainToInstance` or `createValidChainObject` approaches?

On closer look, I believe the reason your responses aren't being instantiated as class instances is because you don't appear to be using the test utilities as shown in other end to end tests. 

Notice in this example: https://github.com/GalaChain/sdk/blob/main/chain-cli/chaincode-template/e2e/lockNFT.spec.ts

```
    // When
    const lockResult = await client.assets.submitTransaction<TokenBalance>(
      "LockTokens",
      lockDto.signed(user1.privateKey),
      TokenBalance
    );
```

Compare to your example we're discussing:

```
    const fetchBalanceRes = await client1.token.FetchBalances(fetchBalanceDto);
    const balanceData = fetchBalanceRes.Data;
    if (balanceData === undefined || balanceData.length === 0) {
      return new BigNumber(0);
    }

    balanceData[0].getQuantityTotal = function (): BigNumber {
      return this.quantity;
    };
    const balanceValue = balanceData[0].getQuantityTotal();

```

You'll need to swithc over to using the `submitTransaction` and `evaluateTransaction` utility methods as designed, rather than calling the contract methods directly.

chain-api/src/types/DexDtos.ts
Divyansh-Gupta-Antier: Since these inputs are expected to accept a wide range of numbers, we have opted not to include additional decorators. Instead, we have simply removed the `@IsNotEmpty` decorator.

chain-api/src/types/DexDtos.ts
Divyansh-Gupta-Antier: Since these inputs are expected to accept a wide range of numbers, we have opted not to include additional decorators. Instead, we have simply removed the `@IsNotEmpty` decorator.

chain-api/src/types/DexV3Pool.ts
Divyansh-Gupta-Antier: We have updated the code to throw more specific errors instead of a generic one.

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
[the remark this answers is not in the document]
Divyansh-Gupta-Antier: In our code, we have utilized the `submitTransaction` and `evaluateTransaction` utility methods. Similar to the apples.spec.ts file, we defined `contractAPI` at the end of the file and invoked it directly in our contract configuration. However, we recently realized that we were not specifying a proper response type while calling these methods, which prevented it from deserializing the response into a specific type. We have addressed this issue in our latest commit, and we can now call the `getQuantityTotal` function directly.

chaincode/src/dex/setProtocolFee.ts
Divyansh-Gupta-Antier: We have implemented the suggested changes in our feeConfig files.

chaincode/src/fees/feeGateImplementations.ts
Divyansh-Gupta-Antier: We have added the required decorators in our contract files.

sentientforest (approved):

PR #526, opened 2025-03-19
Feat: Add package consistency verification

dzikowski:
- Introduced a new function "verifyPackageConsistency" to check for conflicting versions of "@grpc/grpc-js" in the package-lock.json.
- Updated "index.ts" to call "verifyPackageConsistency" during initialization.
- Added "verifyPackageConsistency.ts" to implement the verification logic, including warnings for version conflicts.


chaincode/src/index.ts
```diff
@@ -30,3 +31,5 @@ export * from "./use";
 export * from "./transfer";
 export * from "./vesting";
 export * from "./dex";
+
+verifyPackageConsistency();
```
sentientforest, line 35: I think its probably fine to run on initialization. If we find in the future it adds overhead on startup we could look into exporting the function in the library and importing it into a script/cli that can be setup to only run in specific CI/CD pipelines.

sentientforest (approved):

PR #536, opened 2025-03-28
Fix change the function call in collectTradingFees to the correct fee configuration

Divyansh-Gupta-Antier:
1. Fixed an incorrect function call from Dex `collectTradingFees` to its Launchpad counterpart.
2. Added the missing test cases associated with this error.


Jehosephat (approved):

Jehosephat (approved):
Need to fix linting, but otherwise this looks ok

PR #540, opened 2025-04-05
docs: add new core concepts documentation section

sentientforest:


docs/concepts/chain-key-design.md
```diff
@@ -0,0 +1,211 @@
...
+## What Are Chain Keys?
+
+Chain Keys are ordered collections of property values that together form a unique composite key for storing state data. They leverage Hyperledger Fabric's native composite key capabilities but provide a structured, type-safe approach through GalaChain's decorators.
+
+## Basic Usage
+
+```typescript
+import { ChainObject, ChainKey } from '@galachain/sdk';
+
+export class PlayerInventoryItem extends ChainObject {
```
dzikowski, line 16: there is also the mandatory static `INDEX_KEY` property which defines a class/category (?) of objects. It should be unique per class across the chaincode

docs/concepts/chain-key-design.md
```diff
@@ -0,0 +1,211 @@
...
+### Time-Based Organization
+
+For time-series data or events:
+
+```typescript
+export class GameEvent extends ChainObject {
+  @ChainKey({ position: 0 }) public eventType: string;
+  @ChainKey({ position: 1 }) public year: number;
+  @ChainKey({ position: 2 }) public month: number;
+  @ChainKey({ position: 3 }) public day: number;
```
dzikowski, line 139: This is actually tricky. For range queries, where you typically get ordered results, you will face an issue with string-like ordering for numbers: like `1, 10, 11, 12, 2, 3` for months. For ordering of dates I think those should be strings prepended by zeroes if needed.

docs/concepts/chain-key-design.md
```diff
@@ -0,0 +1,211 @@
...
+   @ChainKey({position: 0}) public readonly type: string;
+   @ChainKey({position: 2}) public readonly id: string; // Gap will cause issues
+   
+   // Correct
+   @ChainKey({position: 0}) public readonly type: string;
+   @ChainKey({position: 1}) public readonly id: string;
+   ```
+   
+   
+2. **Assigning Mutable properties as Chain Key Values**: Avoid using properties that change as chain keys. Changing a chain key would effectively represent a new entity.
```
dzikowski, line 196: also: if you need change one of key properties, you need to delete the old object

docs/concepts/chain-key-design.md
```diff
@@ -0,0 +1,211 @@
...
+
+4. **Using Non-Deterministic Values**: Timestamps or random values make poor chain keys. However, placing values like these at the end of a sequence of chain keys can be a useful strategy to ensure data integrity and prevent collisions.
```
dzikowski, line 198: I know that's not the case here, but that's a tricky thing to mention "random" and "timestamp" words here. Maybe it deserves a mention that any non-deterministic values cannot be created by the chaincode.

docs/concepts/create-read-update-delete.md
```diff
@@ -0,0 +1,127 @@
...
+
+## The GalaChainStubCache
+
+GalaChain implements a transactional cache through the `GalaChainStub` class. This cache ensures:
+
+1. **Read Consistency**: Once an object is read within a transaction, subsequent reads return the same value unless explicitly modified.
+
+2. **Write Atomicity**: All changes are queued and only written to the World State when the transaction successfully completes.
+
+3. **Performance**: Frequently accessed objects are cached to reduce ledger reads.
```
dzikowski, line 94: maybe worth adding something like:

This behavior is different than World State reads in Hyperledger Fabric, when you always get the value that is saved on chain, and in-transaction changes are not visible. (poor words and grammar probably, but you get the idea 😉 )

docs/concepts/identities-wallets-accounts.md
```diff
@@ -0,0 +1,141 @@
...
+## Core Concepts
+
+### User Identities
+
+In GalaChain, users can be identified in multiple ways:
+
+1. **User Alias**: A unique identifier in the format `scheme|value`
+   ```typescript
+   // Examples
+   "eth|0x123...def"  // Ethereum address
```
dzikowski, line 16: always good to stress that this is an address with crc (some chars upper-case)

docs/concepts/identities-wallets-accounts.md
```diff
@@ -0,0 +1,141 @@
...
+- Additional metadata
+
+```typescript
+interface UserProfile {
+  alias: UserAlias;        // Primary identifier
+  ethAddress?: string;     // Optional Ethereum address
+  roles: string[];         // Assigned roles
+  publicKey: string;       // Current public key
+  // ... other fields
+}
```
dzikowski, line 108: actually it is:
```ts
interface UserProfile {
  alias: UserAlias;
  ethAddress?: string;
  tonAddress?: string; // but either eth or ton address is required
  roles?: string[];
}
```

And there is acompannying `PublicKey`:

```
interface PublicKey {
  publicKey: string;
  public signing?: SigningScheme; // "ETH" or "TON"
}
```

Which I'm not sure is a good design to be honest. This is somehow related with the fact that we need both mappings: address -> UserProfile, and alias -> PublicKey, and it was not feasible to perform the migration to keep the state consistent. The more I think about it, the more I'm leaning towards having `publicKey` and `signing` in UserProfile, and instead of `PublicKey`, store just a reference alias -> ethAddress (which allows to get the profile).

docs/concepts/mvcc-read-conflicts.md
```diff
@@ -0,0 +1,100 @@
+# MVCC_READ_CONFLICT Errors
+
+## Overview
+
+MVCC_READ_CONFLICT errors in Hyperledger Fabric occur during transaction validation when multiple transactions attempt to modify the same state concurrently. These errors are a consequence of Fabric's Multi-Version Concurrency Control (MVCC) system, which ensures that changes applied to the ledger remain consistent across the network.
```
dzikowski, line 5: worth to mention that there is also `PHANTOM_READ_CONFLICT`, which is basically the same, but for range reads

docs/concepts/smart-contracts-chaincode.md
```diff
@@ -0,0 +1,198 @@
...
+## Implementation Strategy
+
+When implementing new chaincode:
+
+1. Start with clear business requirements
+2. Design your data model using Chain Objects
+3. Create appropriate DTOs for input/output
+4. Implement contract methods with proper decorators
+5. Add comprehensive unit tests
+6. Document all public interfaces
```
dzikowski, line 198: 7. Remember it's very hard to migrate data once it is saved on chain 😉

dzikowski (approved):

docs/concepts/chain-key-design.md
sentientforest: Could call out, I added some changes and clarifying comments for this one

PR #543, opened 2025-04-10
docs: new Q&A questions and answers section

sentientforest:


docs/questions/burn-tokens-usage.md
```diff
@@ -0,0 +1,97 @@
...
+1. The `TokenBurn` Class:
+GalaChain provides a built-in `TokenBurn` class in `@gala-chain/api` that tracks token burns:
+
+```typescript
+import { TokenBurn } from '@gala-chain/api';
+
+// TokenBurn class structure
+export class TokenBurn extends ChainObject {
+  @ChainKey({ position: 0 })
+  public burnedBy: UserAlias;  // Who performed the burn
```
dzikowski, line 21: I keep wondering about that static `INDEX_KEY` field. Perhaps it would be good to add it to all samples and indicate that this is the first part of the key on chain (perhaps AI could handle that?). It is useful when you inspect those objects in explorer

docs/questions/burn-tokens-usage.md
```diff
@@ -0,0 +1,97 @@
...
+  });
+}
+```
+
+Best Practices:
+- Always validate token ownership before burning
+- Use sequential operations for multiple burns
+- Handle errors gracefully
+- Track burn history
+- Emit burn events
```
dzikowski, line 83: I think that's not accurate

docs/questions/burn-tokens-usage.md
```diff
@@ -0,0 +1,97 @@
...
+
+Key Points:
+- Burns are permanent and irreversible
+- Burns require proper authorization
+- Burns update token balances
+- Burns create audit records
```
dzikowski, line 89: I'm not sure about this statement. Is it related with burn counter? or is it about the transaction history?

docs/questions/burning-NFTs-vs-fungibles.md
```diff
@@ -0,0 +1,120 @@
...
+```typescript
+import { BurnTokenQuantity, TokenInstanceKey } from '@gala-chain/api';
+
+// For NFTs - quantity is always 1
+const nftBurn: BurnTokenQuantity = {
+  tokenInstanceKey: new TokenInstanceKey({
+    collection: 'my-nfts',
+    category: 'characters',
+    type: 'hero',
+    additionalKey: '',
```
dzikowski, line 54: afair typically we use `none`. Empty string may not pass the validation. Also the we don't have this constructor. For NFT though we have: `TokenInstanceKey.nftKey({ collection, cat, type, addK }, instance)`

```suggestion
    additionalKey: 'none',
```

docs/questions/burning-NFTs-vs-fungibles.md
```diff
@@ -0,0 +1,120 @@
...
+
+// For fungible tokens - quantity can be any positive number
+const fungibleBurn: BurnTokenQuantity = {
+  tokenInstanceKey: new TokenInstanceKey({
+    collection: 'game-currency',
+    category: 'coins',
+    type: 'gold',
+    additionalKey: '',
+    instance: '0'  // Usually '0' for fungibles
+  }),
```
dzikowski, line 68:
```suggestion
  tokenInstanceKey: TokenInstanceKey.fungibleKey({
    collection: 'game-currency',
    category: 'coins',
    type: 'gold',
    additionalKey: ''
  }),
```

also a comment `Usually '0' for fungibles` is not accurate, because it's strict - `0` means it is for fungibles

docs/questions/caching-in-chaincode.md
```diff
@@ -0,0 +1,54 @@
...
+   - Cache is automatically flushed at transaction end
+   - All updates are written together
+
+4. Benefits:
+   - Improved performance for repeated reads
+   - Consistent view of state within transaction
+   - Reduced world state database load
+   - Automatic cache management
+
+Note: The stub cache is specifically designed for transaction-level caching. For cross-transaction data access patterns, consider implementing appropriate data structures and query patterns in your chaincode.
```
dzikowski, line 54: also it's worth emphasizing that this is the different behavior that is supported by Fabric by default.

also we don't support cache for range reads (yet)

docs/questions/composite-keys-for-complex-queries.md
```diff
@@ -0,0 +1,100 @@
...
+### Answer
+
+
+Composite keys in GalaChain enable efficient querying by combining multiple fields into a single key. Here's how to use them effectively:
+
+1. Design your composite key structure:
+```typescript
+export class TokenBalance extends ChainObject {
+  @Exclude()
+  public static readonly INDEX_KEY = 'TOKEN_BALANCE';
```
dzikowski, line 16:
```suggestion
  public static readonly INDEX_KEY = 'GCTB';
```

docs/questions/composite-keys-for-complex-queries.md
```diff
@@ -0,0 +1,100 @@
...
+- Progressive filtering: Add more key parts for more specific queries
+- Performance: Composite keys enable fast range queries and prefix scans
+- Uniqueness: Complete composite keys must be unique within a class
+
+Best practices:
+- Order chain keys from most to least frequently queried
+- Use meaningful key parts that support your query patterns
+- Keep key parts short to minimize storage overhead
+- Consider index size impact when designing composite keys
+- Document the composite key structure in class comments
```
dzikowski, line 100: Also: don't use non-primitive types as parts of the key. The only advanced types we support as a part of the keys is `BigNumber` from `bignumber.js` library

Additional thought:

We often use pattern:

```ts
class Obj {
  owner // key0
  collection // key1
  category // key2
  type // key3
  additionalKey // key4
  ...
}
```

May at some point we could support `TokenClassKey` and `TokenInstanceKey` as a part of the key, and the following code:

```ts
class Obj {
  owner // key0
  tokenClass // key1
  ...
}
```

Could produce the same key: `owner\u0collection\u0category\u0type\u0additionalKey`. But that's probably a bit complicated since there is a lot of potential backward compatibility issues

docs/questions/concurrent-operations-in-chaincode.md
```diff
@@ -0,0 +1,230 @@
+### Question
+
+
+How do I implement concurrent operations in chaincode?
```
dzikowski, line 4: I'm not sure if I understand this document in general. Is it AI hallucination? Seems pretty detached from the constraints of MVCC in Fabric. Esp:

---

> 1. Version-Based Concurrency:

I don't understand the benefit and relation to concurrent operations. Seems like a low-level explanation of internal concurrent writes checks / data consistency. Non-relevant I think. Fabric will throw MVCC anyway in case of conflict. Maybe it might be a bit relevant for distributed transactions or sagas, but I don't think we need to cover that in docs.

---

> 2. Atomic Operations:

That probably deserves a dedicated document, explaining that each contract method call is in fact a transaction (and fabric takes care of the data consistency (MVCC)), and about our error handling mechanism (no partial success, all writes flushed only for successful transactions - which is also a big benefit of our cache layer)

---

also 3 and 4 seem a bit low-level and non-relevant

sentientforest: ` Is it AI hallucination? ` - yeah, looks like it. While one could technically write chaincode that way, there wouldn't be any advantage to it or need for it. I rewrote the whole thing to describe concepts and link to other docs.

docs/questions/emit-and-handle-events-in-chaincode.md
```diff
@@ -0,0 +1,130 @@
+### Question
+
+
+What's the best way to emit and handle events in chaincode?
```
dzikowski, line 4: maybe we don't need this file at all?

docs/questions/error-recovery-best-practices.md
```diff
@@ -0,0 +1,95 @@
+### Question
+
+
+What are the best practices for error recovery in chaincode?
```
dzikowski, line 4: I have a feeling both files: `docs/questions/error-recovery-best-practices.md` and `docs/questions/errors-and-exceptions-in-chaincode.md` are mostly irrelevant and misleading. What do we need is:

1. The explanation of our error classes in general (the hierarchy, what is error key, what is error code, common methods). I'm not sure if Q&A is a good place, maybe it would be better for concepts. https://github.com/GalaChain/sdk/blob/main/chain-api/src/utils/error.ts
2. The explanation how errors handling should look like in the transaction - uncaught throw => transaction fails, nothing is saved, and it it guaranteed here: https://github.com/GalaChain/sdk/blob/main/chaincode/src/contracts/GalaTransaction.ts#L232
3. Edge cases, when partial success is allowed, like in `BatchSubmit`

PR #561, opened 2025-05-08
Fix: Remove wrong validation of x509 strings

dzikowski:


chaincode/src/contracts/getCaIdentityAlias.ts
```diff
@@ -38,12 +39,6 @@ export function getCaIdentityAlias(ctx: Context): UserAlias {
     throw new Error("Invalid client account ID format");
```
dviejokfs, line 39: Add the detail of the clientid to add more information to the error

dviejokfs (changes requested):

dviejokfs (approved):

PR #564, opened 2025-05-11
Refactor Dex pool by seperating tick data and position into their own chain objects

Divyansh-Gupta-Antier:


chain-api/src/types/DexDtos.ts
```diff
@@ -722,31 +695,24 @@ export class AddLiquidityResDto extends ChainCallDTO {
   @Type(() => UserBalanceResDto)
   userBalanceDelta: UserBalanceResDto;
 
-  @BigNumberArrayProperty()
-  amounts: BigNumber[];
+  amounts: string[];
```
sentientforest, line 698: This property appears to be missing `class-validator` decorators that reflect the updated type change

chain-api/src/types/DexDtos.ts
```diff
@@ -722,31 +695,24 @@ export class AddLiquidityResDto extends ChainCallDTO {
...
-export class PositionsObject {
-  [key: string]: IPosition[];
-}
-
 export class GetUserPositionsResDto extends ChainCallDTO {
   @IsNotEmpty()
-  @ValidateNested()
-  @Type(() => PositionsObject)
-  positions: PositionsObject;
+  positions: IPosition[];
```
sentientforest, line 709: Currently this `IPosition[]` property is only decorated with `@IsNotEmpty()`. Suggest adding at least one more decorator that is more specific. 

I understand if `IPosition` references an interface, not a Type, it might not work as easily with some of `class-validator`'s options. 

But at a minimum I think we could use something like `@ArrayNotEmpty()` or `@ArrayMinSize(0)` here in addition to `@IsNotEmpty()`.

chain-api/src/types/DexPositionData.ts
```diff
@@ -0,0 +1,152 @@
...
+  @Max(TickData.MAX_TICK)
+  tickUpper: number;
+
+  @ChainKey({ position: 2 })
+  @IsInt()
+  @Min(TickData.MIN_TICK)
+  @IsLessThan("tickUpper")
+  tickLower: number;
+
+  @ChainKey({ position: 4 })
```
sentientforest, line 52:
```suggestion
  @ChainKey({ position: 2 })
  @IsInt()
  @Min(TickData.MIN_TICK)
  @IsLessThan("tickUpper")
  tickLower: number;

  @ChainKey({ position: 3 })
```

Keep the `@ChainKey` positions sequential

sentientforest (approved):
Good work. The new data structures and reduction in complexity of the `Pool` class defined in `DexV3Pool` are steps in the right direction.

chain-api/src/types/DexDtos.ts
Divyansh-Gupta-Antier: Added validations to this property in the latest commit.

chain-api/src/types/DexDtos.ts
Divyansh-Gupta-Antier: Added validations to this property in the latest commit.

chain-api/src/types/DexPositionData.ts
Divyansh-Gupta-Antier: Fixed the skipped chainkey position index.

chain-api/src/types/DexPositionOwner.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  @ChainKey({ position: 1 })
+  @IsNotEmpty()
+  @IsString()
+  poolHash: string;
+
+  @JSONSchema({
+    description: `A tick range mapping that maps a tick range (eg. 10-20) to a unique position ID for this pool`
+  })
+  tickRangeMap: Record<string, string[]>;
```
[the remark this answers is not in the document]
Divyansh-Gupta-Antier: Due to the lack of native support in class-validator for validating `Record` data structures, we’ve implemented a custom validator named `IsStringArrayRecord`. This custom validation has been applied to the corresponding property to ensure its structure adheres to expected constraints.

PR #575, opened 2025-05-16
fix: BatchSubmit does not preserve writes from failed transactions

sentientforest:
// Given
Batch Requests 1, 2, 3

// When
BatchRequest1:

read keyA
write keyA

successful

BatchRequest2:

write keyA <-- gets BatchRequest1's changes

error/failure

BatchRequest3:

read keyA <-- should NOT get BatchRequest2's changes

// Then
BatchRequest3 should not get BatchRequest2's changes

### DIFF
--- a/chaincode/src/types/GalaChainStub.ts
+++ b/chaincode/src/types/GalaChainStub.ts
@@ -135,11 +135,23 @@ class StubCache {
   }
 
   getReads(): Record<string, Uint8Array> {
-    return { ...this.reads };
+    const reads = {};
+
+    for (const [key, value] of Object.entries(this.reads)) {
+      reads[key] = Buffer.from(value.toString());
+    }
+
+    return reads;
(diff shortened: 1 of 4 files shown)


chaincode/src/types/GalaChainStub.ts
```diff
@@ -135,27 +135,62 @@ class StubCache {
...
+    // will silently fail.
+    for (const key of Object.keys(this.writes)) {
+      delete this.writes[key];
+    }
+
+    // deleting all prior keys, and then re-assigning all new key/values
+    // fully updates the existing object with the new object references.
+    for (const key of Object.keys(writes)) {
+      this.writes[key] = writes[key];
+    }
```
sentientforest, line 183: This issue was quite tricky to uncover

Jehosephat (approved):

PR #579, opened 2025-05-19
Add optional positionID field to all Dex operations

Divyansh-Gupta-Antier:


chain-api/src/types/DexDtos.ts
```diff
@@ -269,7 +272,8 @@ export class BurnDto extends ChainCallDTO {
     tickLower: number,
     tickUpper: number,
     amount0Min: BigNumber,
-    amount1Min: BigNumber
+    amount1Min: BigNumber,
+    positionId?: string
```
dzikowski, line 276: this is a low-level constructor, so I'd recommend to pass all parameters explicitly:

```suggestion
    positionId: string | undefined
```

the property declaration should remain as is:

```
public positionId?: string;
```

chain-api/src/types/DexDtos.ts
```diff
@@ -509,14 +520,19 @@ export class CollectDto extends ChainCallDTO {
...
   constructor(
     token0: TokenClassKey,
     token1: TokenClassKey,
     fee: DexFeePercentageTypes,
     amount0Requested: BigNumber,
     amount1Requested: BigNumber,
     tickLower: number,
-    tickUpper: number
+    tickUpper: number,
+    positionId?: string
```
dzikowski, line 535:
```suggestion
    positionId: string | undefined
```

chain-api/src/types/DexDtos.ts
```diff
@@ -579,7 +600,8 @@ export class AddLiquidityDTO extends ChainCallDTO {
     amount0Desired: BigNumber,
     amount1Desired: BigNumber,
     amount0Min: BigNumber,
-    amount1Min: BigNumber
+    amount1Min: BigNumber,
+    positionId?: string
```
dzikowski, line 604:
```suggestion
    positionId: string | undefined
```

chain-api/src/types/DexDtos.ts
```diff
@@ -874,14 +897,19 @@ export class BurnEstimateDto extends ChainCallDTO {
...
   constructor(
     token0: TokenClassKey,
     token1: TokenClassKey,
     fee: DexFeePercentageTypes,
     amount: BigNumber,
     tickLower: number,
     tickUpper: number,
-    owner: string
+    owner: string,
+    positionId?: string
```
dzikowski, line 912:
```suggestion
    positionId: string | undefined
```

chaincode/src/dex/burn.ts
```diff
@@ -78,7 +84,7 @@ export async function burn(ctx: GalaChainContext, dto: BurnDto): Promise<UserBal
     const poolTokenBalance = await fetchOrCreateBalance(ctx, poolAlias, tokenInstanceKeys[index]);
     const roundedAmount = roundTokenAmount(amount, tokenDecimals[index]);
 
-    if (!roundedAmount.isGreaterThan(poolTokenBalance.getQuantityTotal())) {
+    if (roundedAmount.isGreaterThan(poolTokenBalance.getQuantityTotal())) {
```
dzikowski, line 87: is this change intentional?

the PR title does not mention it, while it seems a significant one

chaincode/src/dex/position.helper.ts
```diff
@@ -98,17 +107,27 @@ export async function fetchUserPositionInTickRange(
   poolHash: string,
   tickUpper: number,
   tickLower: number,
+  positionId?: string,
```
dzikowski, line 110:
```suggestion
  positionId: string | undefined,
```

chaincode/src/dex/burn.ts
Divyansh-Gupta-Antier: Yes, this change is intentional. The logic in that code block was flawed, as pointed out in the Certik audit. Previously, the condition was triggered when the amount of tokens the user would receive _was less_ than the pool’s liquidity. However, it should have been triggered when the amount exceeds the available liquidity. This update corrects that logic.

PR #590, opened 2025-05-23
Launchpad reverse bonding curve fee

mistval:
Implements the reverse bonding curve fee feature, which allows token creators to require an extra fee (in native tokens) be paid when selling a launchpad token. The fee increases as the token gets closer to graduation.


Jehosephat (approved):

chain-cli/chaincode-template/e2e/README.md
```diff
@@ -0,0 +1,19 @@
+# Chaincode-template E2E Tests
```
dzikowski, line 1: to be honest I'm not sure if this is the proper file for this part of documentation. Chaincode template serves as a template for new chaincodes. When you run `galachain init <some-dir>`, it creates a directory with the content of `chaincode-template` directory. In the current way it will create a README file with this file. Once you init the chaincode, all required versions were already published, and you don't need `../../npm-pack-and-replace.sh` - you don't even have access to it.

I'm not sure what's the proper location of it, but the instruction here refers to "improving the template" rather than "working on your chaincode". It's for SDK maintainer, and not for custom chaincode developer.

I like this part of docs, and I think it's very useful. I just don't know where to put it, and I don't know if this is a good place.

You may treat it as a week suggestion - we can do something with it later.

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
```diff
@@ -1801,6 +1807,386 @@ describe("LaunchpadContract", () => {
...
+        rbcConfig
+      );
+
+      saleDTO.websiteUrl = "www.rbctest.com";
+      saleDTO.sign(user.privateKey);
+
+      const saleResponse = await client.Launchpad.CreateSale(saleDTO);
+      expect(saleResponse).toEqual(transactionSuccess());
+      const saleVaultAddress = saleResponse.Data?.vaultAddress;
+      assert(saleVaultAddress, "Sale vault address is undefined");
```
dzikowski, line 1949: `expect`?

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
```diff
@@ -1801,6 +1807,386 @@ describe("LaunchpadContract", () => {
...
+
+      const rbcConfig = new ReverseBondingCurveConfigurationDto();
+      rbcConfig.minFeePortion = new BigNumber("0");
+      rbcConfig.maxFeePortion = new BigNumber("0.5");
+
+      const saleDTO = new CreateTokenSaleDTO(
+        "MaxFeeTestA",
+        "MFTA",
+        "Max Fee Test Token",
+        "www.test.com",
```
dzikowski, line 2028: fun fact, you can buy this domain for $4M 😉

npm-pack-and-replace.sh
```diff
@@ -50,7 +50,11 @@ replace_in_pwd_package_json() {
   echo "Packing and replacing $package..."
 
   # pack the package
-  tar_file="$( (cd "$path" && npm run build && npm pack --pack-destination="$current_dir") | tail -n 1)"
+  pushd "$path"
+  npm run build
+  pack_out=$(npm pack --pack-destination="$current_dir")
+  tar_file=$(tail -n 1 <<<"$pack_out")
+  popd
```
dzikowski, line 57: to be honest I like the subconsole approach more, because here once npm run build exits with error, the pwd in current terminal remains changed. If something fails inside subconsole `( ... && ... )`, it doesn't change the pwd

chaincode/src/launchpad/fees.ts
```diff
@@ -0,0 +1,107 @@
...
+    throw new SlippageToleranceExceededError("Fee exceeds maximum acceptable amount");
+  }
+
+  const launchpadConfigKey = ctx.stub.createCompositeKey(LaunchpadFeeConfig.INDEX_KEY, []);
+  const launchpadConfig = await getObjectByKey(ctx, LaunchpadFeeConfig, launchpadConfigKey);
+  const nativeToken = sale.fetchNativeTokenInstanceKey();
+  const { year, month, day } = txUnixTimeToDateIndexKeys(ctx.txUnixTime);
+  const txId = ctx.stub.getTxID();
+
+  await Promise.all([
```
dzikowski, line 77: it's better to avoid `Promise.all` in chaincodes unless you are 100% sure you need it. In Fabric chaincodes are executed by multiple peers, and it needs to be 100% deterministic - otherwise transactions may fail. If for some reason the order of execution matters, and leads to slightly different writes, you should avoid parallel calls.

Also the optimization you get is only for reads, because writes are saved to cache with no IO required, and flushed at the end of the transaction.

dzikowski (approved):

npm-pack-and-replace.sh
mistval: What terminal are you using and how are you running it? I would expect it to leave your cwd alone no matter what the script does. I don't seem to have that issue when I run `../../npm-pack-and-replace.sh` in zsh, my terminal is still in `chain-cli/chaincode-template` even if `npm run build` fails there.

chain-cli/chaincode-template/e2e/launchpad-api.spec.ts
mistval: This assertion is here for type narrowing, otherwise the following line `buyDTO.vaultAddress = saleVaultAddress;` has an error because `Type 'string | undefined' is not assignable to type 'string'`.

chain-cli/chaincode-template/e2e/README.md
mistval: I see, hmm that might be confusing if this README ended up in new projects created with the CLI. I'll just delete this for now, we can dig it back up from commit history if we find a better place for it, and we intend to remove the launchpad code out of the project eventually anyhow.

chaincode/src/launchpad/fees.ts
mistval: Thanks, that makes sense. I pushed a new commit to do them sequentially.

npm-pack-and-replace.sh
dzikowski: interesting, after verifying that it works as expected (the path remains as it should be). So sorry for that, you may ignore my prev comment

PR #621, opened 2025-06-27
Custom batch submit

Jehosephat:


chain-api/src/types/BatchSubmitDtos.ts
```diff
@@ -0,0 +1,59 @@
...
+@JSONSchema({
+  description: "DTO for authorizing users to call BatchSubmit"
+})
+export class AuthorizeBatchSubmitterDto extends ChainCallDTO {
+  @JSONSchema({
+    description: "User to authorize for BatchSubmit operations"
+  })
+  @IsString()
+  @ArrayNotEmpty()
+  authorizers: string[];
```
mistval, line 29: `authorities` would make more sense to me here.

chain-api/src/types/BatchSubmitDtos.ts
```diff
@@ -0,0 +1,59 @@
...
+
+@JSONSchema({
+  description: "DTO for deauthorizing users from calling BatchSubmit"
+})
+export class DeauthorizeBatchSubmitterDto extends ChainCallDTO {
+  @JSONSchema({
+    description: "User to deauthorize from BatchSubmit operations"
+  })
+  @IsString()
+  authorizer: string;
```
mistval, line 40: Similarly, `authority`

chaincode/src/contracts/DexV3Contract.ts
```diff
@@ -81,14 +96,71 @@ import {
...
+      // Use sandboxed context to avoid flushes of writes and deletes, and populate
+      // the stub with current writes and deletes.
+      const sandboxCtx = ctx.createReadOnlyContext(index);
+      sandboxCtx.stub.setWrites(ctx.stub.getWrites());
+      sandboxCtx.stub.setDeletes(ctx.stub.getDeletes());
+
+      // Execute the operation. Collect both successful and failed responses.
+      let response: GalaChainResponse<unknown>;
+      try {
+        if (writesCount >= writesLimit) {
```
mistval, line 142: I usually expect that I'm allowed to hit a "limit", but not exceed it, so consider using strictly greater than.

chaincode/src/dex/batchSubmitAuthorizations.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  try {
+    return await getObjectByKey(ctx, BatchSubmitAuthorizations, key);
+  } catch (error) {
+    // If not found, create a new authorization object with the calling user as the first authority
+    const defaultAuthorizations = new BatchSubmitAuthorizations([ctx.callingUser]);
+    await putChainObject(ctx, defaultAuthorizations);
+    return defaultAuthorizations;
+  }
+}
```
mistval, line 45: Looks like anyone is allowed to call this function, and whoever calls it first becomes the first authorized user. Is that okay? I'd be afraid we forget to call this and someone else does.

Also is there any possible error other than "non found"? Since this overwrites authorities when there's any sort of error, I'd be concerned if there are other possible failures since they could lead to our designated authorities being replaced.

chaincode/src/dex/batchSubmitAuthorizations.ts
```diff
@@ -0,0 +1,125 @@
...
+} from "@gala-chain/api";
+
+import { GalaChainContext } from "../types";
+import { getObjectByKey, putChainObject } from "../utils";
+
+/**
+ * Fetches the batch submit authorizations from the chain.
+ * Creates a new authorization object if none exists.
+ */
+export async function fetchBatchSubmitAuthorizations(
```
sentientforest, line 32: Given the name, I would expect this to be a read-only function, and return an array (because authorizations plural).  Consider renaming the singular chain object `BatchSubmitAuthorizations` to something like `BatchSubmitAuthorizationGroup` (I like group because it reminds me of Linux/Unix user groups used for privliges, but `List`, `Identities`, etc. could work as well)

chaincode/src/dex/batchSubmitAuthorizations.ts
Jehosephat: Yeah, that was a cursor addition that I missed. Totally pulling out any writing that method does, should just be a fetch.

sentientforest (approved):

PR #630, opened 2025-07-02
docs: generate TypeDoc annotations for chain-connect

sentientforest:


sentientforest:
I haven't worked in this module much and I think it could be improved with some TypeDoc annotations and better test coverage. Getting a start on that here.

PR #664, opened 2025-09-04
Draft: Multisig

IndiaJonathan:


chain-cli/network/connection-profiles/cpp-curator.json
```diff
@@ -0,0 +1,24 @@
+{
```
dzikowski, line 1: connection profiled are generated files by SDK and should not be committed

chain-connect/src/customClients/BrowserConnectClient.ts
```diff
@@ -12,7 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { serialize } from "@gala-chain/api";
+import { SignatureDto, SigningScheme, serialize, signatures } from "@gala-chain/api";
```
dzikowski, line 15: I'm not sure we want to add multisig feature to connect library yet -- too many things at once

chain-api/src/types/dtos.ts
```diff
@@ -252,25 +306,47 @@ export class ChainCallDTO {
   }
 
   public sign(privateKey: string, useDer = false): void {
```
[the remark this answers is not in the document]
dzikowski: this way we also can make it more generic and allow multiple signing schemes (ETH/TON)

chain-api/src/types/dtos.ts
```diff
@@ -283,15 +359,66 @@ export class ChainCallDTO {
     return copied;
   }
 
-  public isSignatureValid(publicKey: string): boolean {
```
dzikowski, line 286: I think it would be far simpler to do: 

```ts
public isSignatureValid(publicKey: string, index: number = 0): boolean {
  const signatureObj: ChainCallDTO | SignatureDto | undefined = this.signatures.?[index] ?? (index == 0 ? this : undefined);
  if (signatureObj === undefined) {
    throw new ValidationError(`No signature at index ${index}`);
```

why:
- less changes in the existing code
- no `let`s and mutations (less tricky to analyze by human)

dzikowski: (and also allows multiple signing schemes)

chain-api/src/types/dtos.spec.ts
```diff
@@ -187,13 +190,61 @@ describe("ChainCallDTO", () => {
...
+
+    expect(dto.signatures).toEqual([
+      {
+        signature: "legacy",
+        signerPublicKey: "pk",
+        signerAddress: undefined,
+        signing: undefined,
+        prefix: undefined
+      }
+    ]);
```
dzikowski, line 257: we should have multiple tests:
1. covering the signing with multiple keys, to be sure the target object structure is correct and consistent
2. covering the validation whether the object has a single signature or signatures array
3. verifying multi signatures

chain-api/src/types/dtos.ts
```diff
@@ -494,11 +648,18 @@ export class RegisterUserDto extends SubmitCallDTO {
...
-   * @description Public secp256k1 key (compact or non-compact, hex or base64).
+   * @description Public secp256k1 keys (compact or non-compact, hex or base64).
    */
-  @JSONSchema({ description: "Public secp256k1 key (compact or non-compact, hex or base64)." })
-  @IsNotEmpty()
-  publicKey: string;
+  @JSONSchema({ description: "Public secp256k1 keys (compact or non-compact, hex or base64)." })
+  @IsNotEmpty({ each: true })
+  @ArrayMinSize(1)
+  publicKeys: string[];
```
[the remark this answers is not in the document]
dzikowski: I would consider doing the same thing as with `signature|signatures` - a single object that can either have `publicKey` or an array of `publicKeys` (not both)

